PR #2, opened 2015-02-14
Add draft config format example

gwicke:
Discuss! ;)

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,37 @@
+{
+	"predef": [
+		"ve",
+
+		"setImmediate",
+
+		"QUnit",
+
+		"Map",
+		"Set"
+	],
+
+	"bitwise": true,
+	"laxbreak": true,
+	"curly": true,
+	"eqeqeq": true,
+	"immed": true,
+	"latedef": true,
+	"newcap": true,
+	"noarg": true,
+	"noempty": true,
+	"nonew": true,
+	"regexp": false,
+	"undef": true,
+	"strict": true,
+	"trailing": true,
+
+	"smarttabs": true,
+	"multistr": true,
+
+	"node": true,
+
+	"nomen": false,
+	"loopfunc": true,
+        "esnext": true
+	//"onevar": true
+}
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,37 +1,32 @@
 {
-	"predef": [
-		"ve",
+    "predef": [
+        "setImmediate",
+        "Map",
+        "Set"
+    ],
 
-		"setImmediate",
+    "bitwise": true,
+    "laxbreak": true,
+    "curly": true,
+    "eqeqeq": true,
+    "immed": true,
+    "latedef": true,
+    "newcap": true,
+    "noarg": true,
+    "noempty": true,
+    "nonew": true,
+    "regexp": false,
+    "undef": true,
+    "strict": true,
+    "trailing": true,
 
-		"QUnit",
+    "smarttabs": true,
+    "multistr": true,
 
-		"Map",
-		"Set"
-	],
+    "node": true,
 
-	"bitwise": true,
-	"laxbreak": true,
-	"curly": true,
-	"eqeqeq": true,
-	"immed": true,
-	"latedef": true,
-	"newcap": true,
-	"noarg": true,
-	"noempty": true,
-	"nonew": true,
-	"regexp": false,
-	"undef": true,
-	"strict": true,
-	"trailing": true,
-
-	"smarttabs": true,
-	"multistr": true,
-
-	"node": true,
-
-	"nomen": false,
-	"loopfunc": true,
-        "esnext": true
-	//"onevar": true
+    "nomen": false,
+    "loopfunc": true,
+    "esnext": true
+    //"onevar": true
 }
--- a/README.md
+++ b/README.md
@@ -50,3 +50,8 @@ services:
     interface: localhost
     # more per-service config settings
 ```
+
+# See also
+- https://github.com/strongloop/strong-agent
+- http://krakenjs.com/ - Focused more on MVC & templating rather than
+    supervision & modules
--- a/README.md
+++ b/README.md
@@ -17,9 +17,36 @@ module.exports = function (options) {
 }
 ```
 
-- config loading: standardize on yaml configs (`config.yaml` and
-    `config.example.yaml`) with a few standard top-level values (`port`,
-    `interface`, `num_workers`)
+- [config loading](#config_loading)
 - logging using bunyan & gelf
 - metric reporting using txstatsd
 - heap dumps
+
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -17,9 +17,36 @@ module.exports = function (options) {
...
+# Statsd metrics reporter
+metrics:
+  statsdHost: localhost:8125
+
+services:
+  - name: someService
+    module: ./lib/server.js
+    port: 12345
+    interface: localhost
+    # more per-service config settings
```
d00rman: An immediate extra candidate seems to be `mount` (or `mount-point` or something like that), so that we can cleanly separate multiple services.

gwicke: That assumes that each service is using http, which IMHO is one level up from what this does.

README.md
```diff
@@ -17,9 +17,36 @@ module.exports = function (options) {
...
+    port: <%= @logstash_port %>
+
+# Statsd metrics reporter
+metrics:
+  statsdHost: localhost:8125
+
+services:
+  - name: someService
+    module: ./lib/server.js
+    port: 12345
```
d00rman: Hm, so we could allow multiple ports to be used by multiple services? Wouldn't that complicate things for supervisoid? While I fully concur this is a desirable feature, maybe just reserve it for the future and force all services to use the same port?

README.md
```diff
@@ -17,9 +17,36 @@ module.exports = function (options) {
...
+  level: info
+  streams:
+  # Use gelf-stream -> logstash
+  - type: gelf
+    host: <%= @logstash_host %>
+    port: <%= @logstash_port %>
+
+# Statsd metrics reporter
+metrics:
+  statsdHost: localhost:8125
```
d00rman: Should we allow a specific service to override this?

README.md
```diff
@@ -17,9 +17,36 @@ module.exports = function (options) {
...
+
+```yaml
+num_workers: 1
+
+# Logger info
+logging:
+  level: info
+  streams:
+  # Use gelf-stream -> logstash
+  - type: gelf
```
d00rman: I really dig the user of arrays here!

README.md
gwicke: Right now supervisord actually doesn't care & just forwards this config.

README.md
gwicke: Do you see a use case for different services in the same process using different statsd instances?

README.md
d00rman: True, but the more we generalise, that much more cases we have to cover. So how about we start with REST API services? Then, provide, e.g. *HTTPS* support, at which point all of the services need to use HTTPS (they would have to share the certificate any way).

d00rman:
A general comment is that I believe we should use different configs for different settings - one for development, one for testing and one for production (or have the same for prod and test). This would ease the development, while at the same time provide good/sane values for production. One such example is logging - during development, it is more likely one would use files and/or stdout, rather than logstash. Not to mention that constructs like `<%= @logstash_host %>` require puppet :8ball:

README.md
gwicke: That's all up to the service really. I would think that we'll eventually end up with a standard http web service wrapper (possibly restbase-based for the spec candy, or simply express), but for now I think it's good to remain flexible.

gwicke:
> A general comment is that I believe we should use different configs for different settings

Agreed. Supervisord already supports the -c flag to load a specific config instead of the default.

.jshintrc
```diff
@@ -0,0 +1,37 @@
+{
+	"predef": [
+		"ve",
+
```
d00rman: Lots of empty lines here. Any reason for that?

.jshintrc
```diff
@@ -0,0 +1,37 @@
...
+	"trailing": true,
+
+	"smarttabs": true,
+	"multistr": true,
+
+	"node": true,
+
+	"nomen": false,
+	"loopfunc": true,
+        "esnext": true
```
d00rman: Wrong indentation?

logger.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+    var res = bindAndChild(newLogger);
+
+    // Avoid recursion if there are bugs in the logging code.
+    var inLogger = false;
+
+    function logUnhandledException (err) {
+        if (!inLogger) {
+            inLogger = true;
+            res('error/restbase/unhandled',  err);
```
d00rman: `restbase` should probably be replaced with the service's actual name.

logger.js
```diff
@@ -0,0 +1,100 @@
...
+
+    // Similarly, log uncaught exceptions. Also, exit.
+    process.on('uncaughtException', function(err) {
+        logUnhandledException(err);
+        process.exit(1);
+    });
+    return res;
+}
+
+// TODO: Use constructor instead?
```
d00rman: `makeLogger()` does some useful things, so I'd leave it like that

statsd.js
```diff
@@ -0,0 +1,57 @@
+"use strict";
+var TXStatsD = require('node-txstatsd');
+
+// StatsD wrapper
+function StatsD(statsdHost, statsdPort) {
+    this.statsd = new TXStatsD({
+        host: statsdHost,
+        port: statsdPort,
+        prefix: 'restbase.routes.',
```
d00rman: `prefix` should be parametrised so that each service can have its prefix. Idem for `suffix` one line below.

d00rman:
> Idea here is that there can be multiple services running in each worker.

I'd rather advocate service separation and put each of them in its own process. One simple case: if there is an uncaught exception in a service, the worker is killed, disrupting the others. In general, service separation would also probably be the way they are actually run in production.

.jshintrc
gwicke: Not really. Should actually remove 've' and 'QUnit' from this list.

.jshintrc
gwicke: Looks like it. I just copied this over without looking at it.

logger.js
gwicke: Agreed. This is just an import, haven't changed any of the logic yet.

logger.js
gwicke: We could do the same stream setup from the constructor though.

statsd.js
gwicke: Yup. That's what the name in the config is (partly) for. Just haven't done that yet.

README.md
gwicke: Yeah, bunyan is pretty flexible in that it supports many backends, even at the same time.

gwicke:
> I'd rather advocate service separation and put each of them in its own process.

We would not use this in production for the reasons you mention, but somebody on a small VM might be interested in running several services in a single process to save memory.

logger.js
d00rman: Excellent point. Let's put that in the constructor then (in another PR, ofc)

PR #9, opened 2015-02-17
Bug fix: Let servisor honour the number of workers to fork

d00rman:
A tiny naming bug prevented servisor from starting the sought number of workers. This commit fixes it. Concretely, if the user provides that number as a command-line argument, it is honoured. Otherwise, the num_workers stanza from the configuration file is used. Furthermore, clients can now set num_workers to 'ncpu' in the config, which causes servisor to start as many workers as there are available CPU units on the machine.

### DIFF
--- a/servisor.js
+++ b/servisor.js
@@ -14,6 +14,7 @@ var cluster = require('cluster');
 var path = require('path');
 var yaml = require('js-yaml');
 var fs = Promise.promisifyAll(require('fs'));
+var os = require('os');
 
 
 var Logger = require('./lib/logger');
@@ -68,7 +69,7 @@ Servisor.prototype._sanitizeConfig = function (conf, options) {
         conf.num_workers = options.num_workers;
     } else if(conf.num_workers === 'ncpu') {
         // use the number of CPUs
-        conf.num_workers = require('os').cpus().length;
+        conf.num_workers = os.cpus().length;
     }
     return conf;
 };
--- a/servisor.js
+++ b/servisor.js
@@ -49,7 +49,7 @@ Servisor.prototype.run = function run (conf) {
         config.metrics.name = name;
         self._metrics = new StatsD(config.metrics);
 
-        if (cluster.isMaster && config.numWorkers > 0) {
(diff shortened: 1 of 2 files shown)


servisor.js
```diff
@@ -57,17 +57,26 @@ Servisor.prototype.run = function run (conf) {
...
+Servisor.prototype._sanitizeConfig = function (conf, options) {
     // TODO: Perform proper validation!
     if (!conf.logging) { conf.logging = {}; }
     if (!conf.metrics) { conf.metrics = {}; }
+    // check the number of workers to run
+    if(options.numWorkers !== -1) {
+        // the number of workers has been supplied
+        // on the command line, so honour that
+        conf.num_workers = options.numWorkers;
+    } else if(conf.num_workers === 'ncpu') {
```
gwicke: The default used to be ncpu. Should we keep that?

servisor.js
```diff
@@ -57,17 +58,26 @@ Servisor.prototype.run = function run (conf) {
...
+Servisor.prototype._sanitizeConfig = function (conf, options) {
     // TODO: Perform proper validation!
     if (!conf.logging) { conf.logging = {}; }
     if (!conf.metrics) { conf.metrics = {}; }
+    // check the number of workers to run
+    if(options.num_workers !== -1) {
+        // the number of workers has been supplied
+        // on the command line, so honour that
+        conf.num_workers = options.num_workers;
+    } else if(conf.num_workers === 'ncpu') {
```
gwicke: Should this just be a plain else, so that we default to ncpu if not otherwise set?

d00rman: Yeah, I was thinking about that too while coding it and I'm not sure we should. Take for example `options.num_workers = -1` and `conf.num_workers = 0`. Here we would not honour the config directive. What I would add here, though, is a check:
```javascript
  } else if(conf.num_workers === 'ncpu' || typeof conf.num_workers !== 'number') {
```

What do you think about it? This makes us more cautious while still config-abiding if a number is present

gwicke: Yeah, that should work.

PR #11, opened 2015-02-17
Rename to service-runner

gwicke:
Per discussion at https://github.com/wikimedia/servisor/issues/3

Last rename for a while, I swear!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,20 +1,20 @@
 {
-  "name": "servisor",
+  "name": "service-runner",
   "version": "0.0.0",
   "description": "Generic nodejs service supervisor / cluster runner",
-  "main": "servisor.js",
+  "main": "service-runner.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -25,9 +25,9 @@
   "author": "Wikimedia service team <services@wikimedia.org>",
   "license": "Apache2",
   "bugs": {
-    "url": "https://github.com/wikimedia/servisor/issues"
+    "url": "https://github.com/wikimedia/service-runner/issues"
```
d00rman: Should we put the phabricator link here right away?

gwicke: It doesn't exist yet, only now opened https://phabricator.wikimedia.org/T89747 & officially I'm supposed to wait for a few hours before actually creating it.

Issue #6, opened 2015-02-16
StatsD improvements


atdt:
* The [DWIM](https://en.wikipedia.org/wiki/DWIM) approach to metric names (that is, quietly normalizing metric names) is a poor idea. It is reasonable to demand that develops be deliberate and thoughtful with metric names, because the metric namespace is shared, and care must be taken to ensure it is orderly. It would be optimal if the interface provided by the library automatically constructed metric names, based on what it knew of the application doing the logging and the type of metric that is being logged. But at minimum it should balk (loudly!) at invalid names.
* Instead of `timing()` and `endTiming()`, consider adopting `time()` and `timeEnd()`, for uniformity with the console API (which [Node.js implements](http://nodejs.org/api/console.html#console_console_timeend_label)).
* Having the timing interface accept an array of metric names confuses the calling convention. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). Let the caller iterate if it needs to.

gwicke:
> It would be optimal if the interface provided by the library automatically constructed metric names

Currently, it automatically constructs prefixes based on the service name. This means that all per-service metrics *are* namespaced.

In RESTBase, we need name normalization as we automatically derive metric names from route patterns, which contain braces and slashes. We could expose the normalization as a separate static method instead of doing it implicitly, so that clients could choose whether they want / need it or not.

> Instead of timing() and endTiming(), consider adopting time() and timeEnd()

We are broadly trying to stay compatible with https://github.com/sivy/node-statsd, with the exception of the `endTiming` utility method. Node's console API isn't used widely, and looks like a poor design for asynchronous timings, as concurrent requests will reset each other's timers. To avoid silently producing incorrect metrics, we should keep the statsd client free of per-request state.

> Having the timing interface accept an array of metric names confuses the calling convention.

This is actually an important performance optimization. Sending one UDP packet per timing is expensive both for statsd & the client. In RESTBase, metrics reporting currently uses about as much cpu time as routing. Once / if we move to statsd, we can leverage its batch submission mode to transparently reduce the cpu time spent on metrics reporting.

If we move out the name normalization as discussed above we could remove the suffix parameter. This would also match the batch API as exposed by https://github.com/sivy/node-statsd, which I think is a good idea.

PR #15, opened 2015-02-19
RFC / breaking change: Clean up statsd wrapper

gwicke:
- Conform to node-statsd interface
- Add static normalizeNames method right on the statsd instance & don't
  normalize names by default
- Support both txstatsd & statsd backends

See also #6.

@atdt, @d00rman, please have a look. Don't merge before we have a corresponding restbase change lined up.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -43,6 +43,12 @@ function makeStatsD(options) {
 
     // Add a static utility method for name normalization
     statsd.normalizeNames = normalizeNames;
+
+    // Also add a small utility to send a delta given a startTime
+    statsd.endTiming = function endTiming(names, startTime, samplingInterval) {
+        return this.timing(names, Date.now() - startTime, samplingInterval);
+    };
+
     return statsd;
 }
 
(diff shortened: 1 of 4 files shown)


gwicke:
Also @arlolra @cscott @subbuss

lib/statsd.js
```diff
@@ -1,9 +1,42 @@
...
-    this.statsd = new TXStatsD({
+var nameCache = {};
+function normalizeName (name) {
+    // See https://github.com/etsy/statsd/issues/110
+    // Only [\w_.-] allowed, with '.' being the hierarchy separator.
+    var res = nameCache[name];
+    if (res) {
+        return res;
+    } else {
+        nameCache[name] = name.replace( /[^\/a-zA-Z0-9\.\-]/g, '-' )
```
d00rman: `_` seems to be missing from the regex

gwicke: The intention is to map `_` to `-`, so that slashes are the only char that produces underscores in the final stat name. This is admittedly a bit optimized for restbase path needs.

PR #16, opened 2015-02-21
Add heap reporting and limiting ability

gwicke:
https://phabricator.wikimedia.org/T89932

### DIFF
--- a/lib/heapwatch.js
+++ b/lib/heapwatch.js
@@ -14,11 +14,13 @@ function HeapWatch (conf, logger, statsd) {
 HeapWatch.prototype.watch = function() {
     var usage = process.memoryUsage();
 
-    // Report memory stats to statsd
+    // Report memory stats to statsd. Use 'timing' (which isn't really
+    // timing-specific at all) instead of 'gauge' to get percentiles &
+    // min/max.
     var statsd = this.statsd;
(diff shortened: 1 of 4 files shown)


gwicke:
Heap graph:
image

lib/heapwatch.js
```diff
@@ -0,0 +1,53 @@
...
+    // Report memory stats to statsd. Use 'timing' (which isn't really
+    // timing-specific at all) instead of 'gauge' to get percentiles &
+    // min/max.
+    var statsd = this.statsd;
+    statsd.timing('heap.rss', usage.rss);
+    statsd.timing('heap.total', usage.heapTotal);
+    statsd.timing('heap.used', usage.heapUsed);
+
+    if (usage.heapUsed > this.limit) {
+        console.log(usage.heapUsed, this.limit);
```
d00rman: Why not use `this.logger.log()` (or remove this entirely since something is going to be logged either way) ?

d00rman:
Awesome work @gwicke ! An extremely useful feature. I'd throw some more logic in it to dynamically adapt to the underlying running system. I'll create a ticket for that.

lib/heapwatch.js
gwicke: Good catch! This sneaked through from testing. Removed in #17.

PR #19, opened 2015-02-23
Use package.json info by default, but allow overriding in config

gwicke:
Pull common preperties like name, version and description from package.json by
default, but let people override it in config.yaml.

The previous config.info setting is still supported for b/c, but is going to
be replaced by the 'package' section.


d00rman:
LGTM, but I'll try it our in the morning to be on the safe side :wink:

gwicke:
@d00rman, makes sense. Goodnight! ;)

PR #21, opened 2015-02-24
Minor fix: Use config.package.name for the service's name

d00rman:
The information used for logging and metrics is now available in `config.package`, not `config.info`


d00rman:
Self-merging as this is a simple fix of #19

PR #23, opened 2015-02-26
Keep reporting the stats even when the heap limit has been exceeded

d00rman:
When a worker exceeds its heap limit, it tries to disconnect, otherwise it is forcibly removed. In either case, no more statistics are being reported, so we cannot know exactly when it has exited and how did the heap size change in the last moments. This PR allows workers to keep reporting their stats until they die even if they have been designated for exclusion.

### DIFF
--- a/lib/heapwatch.js
+++ b/lib/heapwatch.js
@@ -11,7 +11,8 @@ function HeapWatch (conf, logger, statsd) {
     this.failCount = 0;
 }
 
-HeapWatch.prototype.watch = function() {
+HeapWatch.prototype.watch = function(doChecklimit) {
+    doChecklimit = doChecklimit === false ? false : true;
     var usage = process.memoryUsage();
 
     // Report memory stats to statsd. Use 'timing' (which isn't really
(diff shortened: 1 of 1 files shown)


gwicke:
@d00rman, as implemented right now the process exits at the latest around the time it would send another stat (60 seconds later), so I think in practice this won't actually make a difference.

I'm more concerned about a DOS effect with the disconnect(). Right now the worker won't accept requests for up to 60 seconds, but also doesn't exit if it has connections open. If that's common this could take the service effectively down for a while. Maybe we should add a mechanism to start a new replacement worker even before the old one actually exits? Risk with that in a DOS / major bug scenario would be spawning an unlimited number of workers, so we'd have to control for that with a check in the master process (ex: max 2x configured number of workers).

d00rman:
> Maybe we should add a mechanism to start a new replacement worker even before the old one actually exits?

Actually, this is a great idea. To improve upon it, the master should control the frequency and total number of spawned (and active) workers. Throttling would prevent us from killing ourselves due to process exhaustion.

So, when a worker reaches the memory threshold (for 3 times consecutively), it sends a message to the master and then calls `disconnect()`. The master aggregates worker messages for, say, 3 seconds, and then spawns the needed workers taking into account their total number (so that the given num_workers config is not surpassed).

> right now the process exits at the latest around the time it would send another stat (60 seconds later)

Why did I think reporting takes place in intervals shorter than that? Ah, ok. Disregard my PR then, but let's keep the discussion about restarts going!

gwicke:
> So, when a worker reaches the memory threshold (for 3 times consecutively), it sends a message to the master and then calls disconnect(). The master aggregates worker messages for, say, 3 seconds, and then spawns the needed workers taking into account their total number (so that the given num_workers config is not surpassed).

Yeah, that sounds good to me.

PR #24, opened 2015-03-23
Stats: Normalise the service name and include it in LogStatsD reports

d00rman:
Users can choose whether to normalise their stats name or not. However, the service name itself, prefixed to all statistics, should be normalised. Also, when sending the stats to the logger, make it log the complete statistic name as it would appear in production.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -18,10 +18,12 @@ function normalizeName (name) {
 
 // A simple console reporter. Useful for development.
 var methods = ['timing','increment','decrement','gauge','unique'];
-function LogStatsD(logger) {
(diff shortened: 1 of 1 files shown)


lib/statsd.js
```diff
@@ -49,7 +52,7 @@ function makeStatsD(options, logger) {
     if (options.type === 'txstatsd') {
         statsd = new TXStatsD(statsdOptions);
     } else if (options.type === 'log') {
-        statsd = new LogStatsD(logger);
+        statsd = new LogStatsD(logger, srvName);
```
gwicke: Alternatively, we could pass in the same statsdOptions here.

PR #29, opened 2015-04-09
Fix txstatsd support & release v0.1.7

gwicke:
f5986f5a1e9 broke txstatsd support, and this patch basically reverts it. The
txtstats option was always ignored by the statsd backend library, but in this
patch we are also setting it to false to make things clearer.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -44,10 +44,10 @@ function makeStatsD(options, logger) {
         port: options.port,
         prefix:  srvName + '.',
         suffix: '',
-        txstatsd  : false,
-        globalize : false,
-        cacheDns  : true,
-        mock      : false
+        txstatsd: options.type === 'txstatsd',
+        globalize: false,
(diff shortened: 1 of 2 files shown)


gwicke:
@d00rman @atdt

lib/statsd.js
```diff
@@ -44,10 +44,10 @@ function makeStatsD(options, logger) {
         port: options.port,
         prefix:  srvName + '.',
         suffix: '',
-        txstatsd  : false,
-        globalize : false,
-        cacheDns  : true,
-        mock      : false
+        txstatsd: options.type === stats,
```
d00rman: I don't understand this. `type === stats` ?

d00rman: What would the other options be? Why not txtstatsd?

gwicke: Sorry, didn't have my morning coffee yet ;)

PR #30, opened 2015-04-17
Auto-generate updates for the deploy repository using Docker

d00rman:
This PR is a port of wikimedia/service-template-node#36 to service-runner. It holds the same functionality, but with different command-line arguments:
- `-s, --docker-start` - starts the service inside a container
- `-t, --docker-test` - starts the test process inside a container
- `-b, --build` - builds/updates the deploy repository
- `-f, --force-build` - forces node module rebuild even if there are no changes in the source repository
- `-r, --review` - build the deploy repository and send the patch for review to Gerrit

Additionally, a small fix has been included that shows the service's version (and not that of service-runner) when the `-v` command-line argument is supplied.

Bug: [T96128](https://phabricator.wikimedia.org/T96128)

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -50,9 +50,6 @@ function promised_spawn(args, capture) {
             child.stdout.on('data', function(data) {
                 ret += data.toString();
             });
-            child.stderr.on('data', function(data) {
-                console.log(data.toString());
-            });
         }
         child.on('close', function() {
             child = undefined;
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -390,22 +390,15 @@ function sig_handle() {
 
 function main(options, configuration) {
 
-    // set up working options
-    if(options.hasOwnProperty('forceBuild')) {
-        // called from service-runner, translate
-        opts = {
-            tests: options.dockerTest,
-            coverage: false,
-            deploy: options.build,
-            verbose: false,
-            need_build: options.forceBuild,
-            review: options.sendReview,
-            path: options.basePath
-        };
-    } else {
-        // called from the command line
-        opts = options;
-    }
+    opts = {
+        tests: options.dockerTest,
+        coverage: false,
+        deploy: options.buildDeploy,
+        verbose: options.verbose,
+        need_build: options.buildDeploy && options.force,
+        review: options.sendReview,
+        path: options.basePath
+    };
 
     // config and package info
     config = configuration;
@@ -457,62 +450,5 @@ function main(options, configuration) {
 }
 
 
-if(module.parent === null) {
-
-    var opts = {
-        tests: false,
-        coverage: false,
-        deploy: false,
-        verbose: false,
-        need_build: false,
-        review: false,
-        path: path.resolve(__dirname, '..')
-    };
-
-    // check for command-line args
-    var args = process.argv.slice(2);
-    var arg;
-    while((arg = args.shift()) !== undefined) {
-        switch(arg) {
-            case '-t':
-            case '--test':
-                opts.tests = true;
-                break;
-            case '-c':
-            case '--cover':
-                opts.coverage = true;
-                break;
-            case '-b':
-            case '--build-deploy':
-                opts.need_build = true;
-                /* falls through */
-            case '-d':
-            case '--update-deploy':
-                opts.deploy = true;
-                break;
-            case '-v':
-            case '--verbose':
-                opts.verbose = true;
-                break;
-            default:
-                console.log('This is a utility script for starting service containers using docker.');
-                console.log('Usage: ' + process.argv.slice(0, 2).join(' ') + ' [OPTIONS]');
-                console.log('Options are:');
-                console.log('  -t, --test           instead of starting the service, run the tests');
-                console.log('  -c, --cover          run the tests and report the coverage info');
-                console.log('  -d, --update-deploy  update the deploy repo');
-                console.log('  -b, --build-deploy   update the deploy repo and force module rebuild');
-                console.log('  -v, --verbose        output the commands being run');
-                process.exit(/^-(h|-help)/.test(arg) ? 0 : 1);
-        }
-    }
-
-    // start the process
-    main(opts);
-
-} else {
-
-    module.exports = main;
-
-}
+module.exports = main;
 
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -1,0 +1,510 @@
+#!/usr/bin/env node
+
+
+'use strict';
+
+var spawn = require('child_process').spawn;
+var yaml = require('js-yaml');
+var P = require('bluebird');
+var fs = P.promisifyAll(require('fs'));
+var path = require('path');
+
+
+// info from the package definition
+var pkg;
+// info from the service-runner config file
+var config;
+// target info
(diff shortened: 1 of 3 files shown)


d00rman:
cc @gwicke @eevans

gwicke:
This PR is a port of wikimedia/service-template-node#36 to service-runner. It holds the same functionality, but with different command-line arguments:
- `-s, --docker-start` - starts the service inside a container
- `-t, --docker-test` - starts the test process inside a container

How about `start-docker` and `test-docker` as positional action parameters, along with `npm start-docker` and `npm test-docker`?

- `-b, --build` - builds/updates the deploy repository

Since we want to build other things too, how about `build deploy-repo`?

- `-f, --force-build` - forces node module rebuild even if there are no changes in the source repository

This could be documented as more orthogonal `--force`, with what is forced depending on the main action.

- `-r, --review` - build the deploy repository and send the patch for review to Gerrit

This is very WMF-specific. Can we document this as an option to the main action (`build deploy-repo` in this case)?

lib/docker.js
```diff
@@ -0,0 +1,518 @@
...
+
+/**
+ * Wraps a child process spawn in a promise which resolves
+ * when the child process exists.
+ *
+ * @param {Array} args the command and its arguments to run (uses /usr/bin/env)
+ * @param {Boolean} capture whether to capture stdout and return its contents
+ * @return {Promise} the promise which is fulfilled once the child exists
+ */
+function promised_spawn(args, capture) {
```
gwicke: You could use execFileAsync instead.

lib/docker.js
```diff
@@ -0,0 +1,518 @@
...
+    var contents = '';
+    var base_img;
+    var extra_pkgs = ['nodejs', 'nodejs-legacy', 'npm', 'git'];
+
+    // set some defaults
+    if(!pkg.deploy) {
+        pkg.deploy = {};
+    }
+    if(!pkg.deploy.target) {
+        pkg.deploy.target = 'ubuntu';
```
gwicke: 'debian'?

lib/docker.js
```diff
@@ -0,0 +1,518 @@
...
+        verbose: false,
+        need_build: false,
+        review: false,
+        path: path.resolve(__dirname, '..')
+    };
+
+    // check for command-line args
+    var args = process.argv.slice(2);
+    var arg;
+    while((arg = args.shift()) !== undefined) {
```
gwicke: You might want to use a parameter parse package like yargs here instead.

lib/docker.js
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    // target info
+    try {
+        targets = yaml.safeLoad(fs.readFileSync(path.join(opts.path, 'targets.yaml')));
+    } catch(e) {
+        // no such file or wrong format,
+        // set the defaults
+        targets = {
+            debian: 'buildpack-deps:jessie',
```
gwicke: Should we use https://registry.hub.docker.com/_/debian/ ? It is [recommended on the Debian wiki](https://wiki.debian.org/Docker). They also have [a review](https://joeyh.name/blog/entry/docker_run_debian/) of the differences between the image and a vanilla debootstrap that sounds reasonable.

service-runner.js
```diff
@@ -238,37 +254,47 @@ ServiceRunner.prototype._getOptions = function (opts) {
             // all concurrent short requests.
             n: -1,
             c: './config.yaml',
+            b: false,
```
gwicke: See my comment on an alternate command-based parameter scheme.

gwicke:
LGTM in principle, just some detail stuff.

d00rman:
> How about start-docker and test-docker as positional action parameters, along with npm start-docker and npm test-docker?
> -b, --build - builds/updates the deploy repository
> Since we want to build other things too, how about build deploy-repo?
> 
> -f, --force-build - forces node module rebuild even if there are no changes in the source repository
> This could be documented as more orthogonal --force, with what is forced depending on the main action.
> 
> -r, --review - build the deploy repository and send the patch for review to Gerrit
> This is very WMF-specific. Can we document this as an option to the main action (build deploy-repo in this case)?

All of these have been addressed in 971fe0e . I put `docker-start` and `docker-test` instead, since `start-docker` and `test-docker` imply (to me, at least) you want to start and test Docker, whereas `docker-start` and `docker-test` somewhat clarify that you want Docker to start and test your service.

d00rman:
> along with npm start-docker and npm test-docker?

I didn't put this one, as I can see its use. Defining npm scripts in `package.json` makes them reachable only from service-runner's package, not its dependants. The service template already has these, I'll just modify them to use service-runner instead of docker.js once this PR is merged, and the new version ihas been published.

lib/docker.js
d00rman: For some executions I want to capture the output altogether (like `git status`), while for others it is good for the user to see the output in real time (like building the image, or `npm install`).

d00rman:
> 'debian'? ;)

> Should we use https://registry.hub.docker.com/_/debian/ ? It is recommended on the Debian wiki. They also have a review of the differences between the image and a vanilla debootstrap that sounds reasonable.

Done in 4e6adde

d00rman:
> Can we make 'build' more universal? We'll want to build other things too, like debian packages, docker images, init scripts etc. build deploy-repo would leave the door open for that.

Not sure I follow. Currently, `service-runner build` is intended (reserved) for building everything, while `service-runner build --deploy-repo` builds only the deploy repo.

d00rman:
> Since the builds are mutually exclusive and have different options,

TBH, I'm still in a pickle whether they should be mutually-exclusive. If so, then I'd be fine with `build-deploy[-repo]` and `build-package` as well. If, on the other hand, `build` alone builds everything there is to build, then options would be more appropriate wrt consistency IMO.

>I still think that subcommands would be cleaner and easier to document.

Heh, I'm not sure that's true, given that `yargs` doesn't support subcommands, so having them would involve having more code based on the chosen command (parser resetting, adding options etc), not to mention the fact that we couldn't profit from its autogenerated docs. I would be in favour of passing totally to commands, as the current dual system seems a bit inconsistent, e.g. `-c` doesn't need a command, while `-r` does. Maybe if we introduced a `run` command and make it the default? (We would still need to build help docs ourselves in that case)

Granted, the current help message could be further enriched and bettered, but it's already a step forward :)

lib/docker.js
```diff
@@ -0,0 +1,454 @@
...
+            return start_container(['-v', opts.dir + ':/opt/service', '-w', '/opt/service']);
+        }).then(function() {
+            // remove unnecessary files
+            return promised_spawn([
+                'find',
+                'node_modules/',
+                '-regextype',
+                'posix-egrep',
+                '-iregex',
+                "(.*\\.git.*|.*\\.md|.*\\.txt|.*readme|.*licence)",
```
gwicke: Afaik the only things we really *need* to remove are `.gitignore` files, as they'll mess with the checkin. Removing the other files you match here is *probably* fine, but I'm not sure that it's worth it.

gwicke:
Merged as-is per discussion this morning. Plan is to beautify argument / command handling in a follow-up.

PR #32, opened 2015-05-13
Start up workers sequentially

gwicke:
Startup tasks like schema migrations or table creations don't necessarily do
well when executed in many workers at once. Fox example, Cassandra does not
like highly concurrent schema changes, and can get into a state where nodes
disagree on the schema version until they are restarted.

This patch addresses that issue by starting up workers sequentially. This
means that schema migrations can finish without concurrency in the first
worker that starts up.

The patch also adds a small random delay for restarts on failure.

Next steps: Also support a rolling restart, not just a rolling startup.

Task: https://phabricator.wikimedia.org/T98936

### DIFF
--- a/service-runner.js
+++ b/service-runner.js
@@ -133,9 +133,8 @@ ServiceRunner.prototype._runMaster = function() {
     this._logger.log('info/service-runner', 'master(' + process.pid + ') initializing '
             + this.config.num_workers + ' workers');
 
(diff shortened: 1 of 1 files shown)


eevans:
Neat!

d00rman:
This patch forces a sequential start-up, thus changing the default behaviour of `service-runner` (without even bumping the version). I would really prefer it to be an argument. Thoughts?

PR #34, opened 2015-05-14
Fix bugs introduced in 0.1.9 & release v0.1.10

gwicke:
- Make sure we return the servers when starting up without forking. This
  somewhat undocumented feature is used by the restbase test runner to access
  the server instance. TODO: Either define that interface properly, or find a
  way to avoid using it.
- Wait for all workers to start up before returning.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "service-runner",
-  "version": "0.1.9",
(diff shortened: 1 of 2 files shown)


d00rman:
Yup, that looks more correct :) And it's used by all service-template-based services too in tests.

PR #35, opened 2015-05-15
Add the capability to perform rolling restarts on SIGHUP

gwicke:
This adds support for zero-downtime rolling worker restarts. With at least two
workers, all services can be reloaded without dropping any requests.

Task: https://phabricator.wikimedia.org/T98936

### DIFF
--- a/service-runner.js
+++ b/service-runner.js
@@ -37,6 +37,8 @@ function ServiceRunner(options) {
     this._logger = null;
     this._metrics = null;
 
+    this._workers = {};
+
     // Figure out the base path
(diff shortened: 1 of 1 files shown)


d00rman:
It doesn't seem to work for me. Not entirely at least, in the sense that new processes get started by the master, but the old ones still hang around. I am unable to tell at this point if they are still active in connection allocation or not, but they are definitely still connected to the master.

I tried to change various things in order to force a proper worker exit, but was unable to do so.

d00rman:
Works now!

PR #37, opened 2015-06-18
Use cluster.workers instead of maintaining our own worker registry

gwicke:
The cluster module already maintains an object indexing the current workers,
so we don't need to maintain one ourselves. This simplifies the code slightly.


gwicke:
This doesn't really make a difference to heap limit enforcement. In local testing that is working before & after.

PR #38, opened 2015-06-20
Don't restart workers if they exited because of a disconnect()

gwicke:


d00rman:
Sorry to react late on this PR, but I don't think this is correct. [lib/heapwatch.js#L35](https://github.com/wikimedia/service-runner/blob/master/lib/heapwatch.js#L35) calls `worker.disconnect()` as well, which sets the `suicide` flag, meaning that if a worker kills itself because of a memory excess, it's not going to be restarted by the master any more (unless 60 seconds pass and `process.exit()` is invoked).

PR #41, opened 2015-07-13
Logger improvements

d00rman:
The logger uses the `_levelMatcher` private property to determine if a log entry should be logged or not. However, the threshold is set based on `conf.level`, without taking into account individual stream levels. That causes log entries for levels lower than the global level not to be logged in the individual log streams.

This PR performs the check to find the minimum log level to use correctly based on individual stream settings. Additionally, it explicitly appends the `level` property to streams lacking it. The stream conversion has been slightly improved to ease the addition of new stream converters, and converters for `stdout` and `stderr` have been added.

Bug: [T105500](https://phabricator.wikimedia.org/T105500)

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -27,6 +27,31 @@ function Logger(confOrLogger, args) {
     this.args = args;
 }
 
+var streamConverter = {
+    gelf: function(stream, conf) {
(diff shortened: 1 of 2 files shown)


d00rman:
cc @gwicke @eevans

PR #43, opened 2015-08-12
Logging audit

Pchelolo:
Self requires the message in the log to be under the field 'msg', but we name it incorrectly sometimes. This fixes it.

Bug: https://phabricator.wikimedia.org/T108189

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -129,23 +129,41 @@ Logger.prototype.child = function(args) {
 
 Logger.prototype.log = function(level, info) {
     var levelMatch = this._levelMatcher.exec(level);
-    if (levelMatch) {
+    if (info && levelMatch) {
         var logger = this._logger;
         var simpleLevel = levelMatch[1];
-        if (info && info instanceof String) {
-            info = {msg: info};
+
+        if (info instanceof String) {
+            // Need to convert to primitive
+            info = info.toString();
         }
-        if (info && typeof info === 'object') {
+
+        if (typeof info === 'string') {
+            logger[simpleLevel].call(logger, info);
+        } else if (typeof info === 'object') {
+            var msg;
             // Got an object
             //
+            // If there's a wrapped Error - unwrap and send,
+            // bunyan will properly handle it
+            if (info.err instanceof Error) {
+                info = info.err;
+            } else if (info.msg || info.message || info.info) {
+                msg = info.msg || info.message || info.info
+            }
+
             // Inject the detailed levelpath.
             // 'level' is already used for the numeric level.
             info.levelPath = level;
 
             // Also pass in default parameters
             info = extend(info, this.args);
+            if (msg) {
+                logger[simpleLevel].call(logger, info, msg);
+            } else {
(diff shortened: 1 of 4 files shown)


lib/logger.js
```diff
@@ -129,23 +129,41 @@ Logger.prototype.child = function(args) {
 
 Logger.prototype.log = function(level, info) {
     var levelMatch = this._levelMatcher.exec(level);
-    if (levelMatch) {
+    if (info && levelMatch) {
         var logger = this._logger;
         var simpleLevel = levelMatch[1];
-        if (info && info instanceof String) {
-            info = {msg: info};
```
Pchelolo: This was a double-bug. `"test" instanceof String` is actually `false`, but we didn't need to wrap a message into an object - it gets overriden by bunyan with a third argument value (or by undefined).

lib/logger.js
```diff
@@ -129,23 +129,41 @@ Logger.prototype.child = function(args) {
...
+        if (typeof info === 'string') {
+            logger[simpleLevel].call(logger, info);
+        } else if (typeof info === 'object') {
+            var msg;
             // Got an object
             //
+            // If there's a wrapped Error - unwrap and send,
+            // bunyan will properly handle it
+            if (info.err instanceof Error) {
+                info = info.err;
```
Pchelolo: Errors are correctly handled by bunyan, no need to change anything about them

service-runner.js
```diff
@@ -69,13 +69,13 @@ ServiceRunner.prototype.run = function run(conf) {
 
         // Set up the logger
         if (!config.logging.name) {
-            config.logging.name = name;
+            config.logging.name = (config.services && config.services.name) || name;
```
d00rman: `config.services` is an Array. But. this is not the right place nor way to set the name. [`_runWorker()`](https://github.com/wikimedia/service-runner/blob/master/service-runner.js#L278-L285) conveys that info to individual services and already sets the name.

service-runner.js
```diff
@@ -69,13 +69,13 @@ ServiceRunner.prototype.run = function run(conf) {
...
         }
         self._logger = new Logger(config.logging);
 
         // And the statsd client
         if (!config.metrics.name) {
-            config.metrics.name = name;
+            config.metrics.name = (config.services && config.services.name) || name;
```
d00rman: Idem as above, with the distinction that `config.metrics` is a shared object, so a new metrics object should be created for each service (in [`_runWorker()`](https://github.com/wikimedia/service-runner/blob/master/service-runner.js#L278-L285))

Pchelolo:
@d00rman Updated according to your comments.

service-runner.js
```diff
@@ -252,6 +251,8 @@ ServiceRunner.prototype._runWorker = function() {
         process.chdir(cwd);
     });
 
+    self._metrics = makeStatsD(self.config.metrics, self._logger);
```
d00rman: Need to add the check for `self.config.metrics.name` (from L#76), and set it to the value of L#56

PR #44, opened 2015-08-14
Implemented heartbeat in service-runner

Pchelolo:
The ticket opens multiple questions, but this PR tackles the solutions with heartbeat from workers, which is tracked by master, and if it stops, the an attempt to make a graceful shutdown is made, if that fails, woke is killed and restarted. 

I've been feeling myself as a serial killer today, killing and breaking workers with different kinds of weapons. The biggest problem is CPU-bound loops. If a worker got in some kind of `while(true)`, lots of weird things start happening:
- Obviously, before the fix we couldn't detect it and had it running forever.
- Killing it with `worker.kill('SIGKILL')` doesn't work, because `worker.kill` doesn't send a signal right away, but a disconnect first, which can't finish for a worker that's in a loop.
- Killing a worker that got to a CPU-intense loop with console `kill -9 <pid>` crashes the whole server!!! (including master). The problem is this bug in node: https://github.com/joyent/node/issues/9409 Master's notion of workers is updated on `disconnect` message and checked on `exit`. This was fixed by a dirty hack which replaces node's listeners on a worker, but the hack works and it looks pretty stable.

I've run the service with the fix for quite some time, performed some benchmarking, killing workers in the process, and everything looks stable. What are your opinions on this? Do we want a dirty hack? 

Bug: https://phabricator.wikimedia.org/T90362

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -3,6 +3,10 @@
 # Use 'ncpu' to run as many workers as there are CPU units
 num_workers: ncpu
 
+# Number of milliseconds to wait for a heartbeat from worker before killing
+# and restarting it
+worker_heartbeat_timeout: 1000
+
 # Log error messages and gracefully restart a worker if v8 reports that it
 # uses more heap (note: not RSS) than this many mb.
 worker_heap_limit_mb: 500
--- a/config.yaml
+++ b/config.yaml
@@ -5,7 +5,7 @@ num_workers: ncpu
 
 # Number of milliseconds to wait for a heartbeat from worker before killing
 # and restarting it
-worker_heartbeat_timeout: 1000
+worker_heartbeat_timeout: 75000
 
 # Log error messages and gracefully restart a worker if v8 reports that it
 # uses more heap (note: not RSS) than this many mb.
--- a/config.yaml
+++ b/config.yaml
@@ -5,7 +5,7 @@ num_workers: ncpu
 
 # Number of milliseconds to wait for a heartbeat from worker before killing
 # and restarting it
-worker_heartbeat_timeout: 75000
+worker_heartbeat_timeout: 7500
 
 # Log error messages and gracefully restart a worker if v8 reports that it
 # uses more heap (note: not RSS) than this many mb.
--- a/service-runner.js
+++ b/service-runner.js
@@ -150,10 +150,11 @@ ServiceRunner.prototype._checkHeartbeat = function() {
     var self = this;
     self.interval = setInterval(function() {
         if (!self._shuttingDown && !self._inRollingRestart) {
+            var now = new Date();
             Object.keys(cluster.workers).forEach(function(workerId) {
                 var worker = cluster.workers[workerId];
                 var lastBeat = self.workerHeartbeatTime[worker.process.pid];
-                if (!lastBeat || (!lastBeat.killed && new Date() - lastBeat.time
+                if (!lastBeat || (!lastBeat.killed && now - lastBeat.time
                         > self.config.worker_heartbeat_timeout)) {
                     self._logger.log('error/service-runner/master',
                         'worker ' + worker.process.pid + ' stopped sending heartbeats, killing.');
--- a/service-runner.js
+++ b/service-runner.js
@@ -46,6 +46,8 @@ function ServiceRunner(options) {
 
     // Are we performing a rolling restart?
     this._inRollingRestart = false;
+
+    this.workerHeartbeatTime = {};
 }
 
 ServiceRunner.prototype.run = function run(conf) {
@@ -100,6 +102,7 @@ ServiceRunner.prototype._sanitizeConfig = function(conf, options) {
         // use the number of CPUs
         conf.num_workers = os.cpus().length;
     }
+    conf.timeout = conf.timeout || 1000;
     return conf;
 };
 
@@ -143,6 +146,24 @@ ServiceRunner.prototype.updateConfig = function updateConfig(conf) {
     }
 };
 
+ServiceRunner.prototype._checkHeartbeat = function() {
+    var self = this;
+    self.interval = setInterval(function() {
+        if (!self._shuttingDown && !self._inRollingRestart) {
+            Object.keys(cluster.workers).forEach(function(workerId) {
+                var worker = cluster.workers[workerId];
+                var lastBeat = self.workerHeartbeatTime[worker.process.pid];
+                if (!lastBeat || (!lastBeat.killed && new Date() - lastBeat.time > 3 * self.config.timeout)) {
+                    self._logger.log('error/service-runner/master',
+                        'worker ' + worker.process.pid + ' stopped sending heartbeats, killing.');
+                    self._stopWorker(workerId);
+                    // Don't need to respawn a worker, it will be restarted upon 'exit' event
+                }
+            });
+        }
+    }, self.config.timeout);
+};
+
 ServiceRunner.prototype._runMaster = function() {
     var self = this;
     // Fork workers.
@@ -153,11 +174,12 @@ ServiceRunner.prototype._runMaster = function() {
         if (!self._shuttingDown && !self._inRollingRestart) {
             var exitCode = worker.process.exitCode;
             self._logger.log('error/service-runner/master',
-                    'worker' + worker.process.pid
-                    + 'died (' + exitCode + '), restarting.');
+                    'worker ' + worker.process.pid
+                    + ' died (' + exitCode + '), restarting.');
+            delete self.workerHeartbeatTime[worker.process.pid];
             P.delay(Math.random() * 2000)
             .then(function() {
-                cluster.fork();
+                self._startWorkers(1);
             });
         }
     });
@@ -166,6 +188,9 @@ ServiceRunner.prototype._runMaster = function() {
         self._shuttingDown = true;
         self._logger.log('info/service-runner/master',
                 'master shutting down, killing workers');
+        if (self.interval) {
+            clearInterval(self.interval);
+        }
         cluster.disconnect(function() {
             self._logger.log('info/service-runner/master', 'Exiting master');
             process.exit(0);
@@ -178,19 +203,35 @@ ServiceRunner.prototype._runMaster = function() {
     // Set up rolling restarts
     process.on('SIGHUP', this._rollingRestart.bind(this));
 
-    return this._startWorkers(this.config.num_workers);
+    return this._startWorkers(this.config.num_workers)
+    .then(function(workers) {
+        setTimeout(function() {
+            self._checkHeartbeat();
+        }, 1000);
+        return workers;
+    });
 };
 
-ServiceRunner.prototype._stopWorker = function(id) {
-    var worker = cluster.workers[id];
+ServiceRunner.prototype._stopWorker = function(workerId) {
+    var self = this;
+    var worker = cluster.workers[workerId];
+    self.workerHeartbeatTime[worker.process.pid] = {
+        time: null,
+        killed: true
+    };
     var res = new P(function(resolve) {
         var timeout = setTimeout(function() {
-            worker.kill('SIGKILL');
+            // worker.kill doesn't send a signal immediately, it waits until
+            // worker closes all connections with master. If after a minute
+            // it didn't happen, don't expect it happen ever.
+            process.kill(worker.process.pid, 'SIGKILL');
+            delete self.workerHeartbeatTime[worker.process.pid];
             resolve();
         }, 60000);
         worker.on('disconnect', function() {
             worker.kill('SIGKILL');
             clearTimeout(timeout);
+            delete self.workerHeartbeatTime[worker.process.pid];
             resolve();
         });
     });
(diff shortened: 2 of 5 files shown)


service-runner.js
```diff
@@ -215,25 +256,71 @@ ServiceRunner.prototype._rollingRestart = function() {
...
+/*
+    This is a workaround for the following bug in node:
+    https://github.com/joyent/node/issues/9409
+
+    Shortly, the problem is that a worker is removed by master
+    if it sends 'disconnect' and then it's checked on 'exit'.
+    However, in some cases (for example an infinite CPU loop in worker,
+    'disconnect' may never happen, and master will crash receiving an 'exit'
+ */
+function workAround(worker) {
```
gwicke: This name doesn't give much information on what this actually does. How about something like `fixExitDisconnectListeners`?

gwicke:
On IRC, @arlolra brought up the idea to also keep track of what each worker is doing by posting a 'starting work on X' message on each new request. However, that is service-dependent & can added later.

Pchelolo:
@gwicke the 'start working on X' can't be used as a heartbeat, because if there's no requests - nothing happens - worker get's recreated for no reason. Also, I'm not sure I understand that idea - what is master going to do in response to 'start working on X' message? I can't imagine how this could be used except logging 'restarting worker due to loss of heartbeat, the last thing it started to do is X'.

gwicke:
@Pchelolo, it would just provide information on what the worker was doing at the time when the timeout triggers. So, it really only changes what information is available for logging, but not the behavior.

In any case, lets consider that separately.

gwicke:
@d00rman, could you give this a final check-over & perhaps merge?

d00rman:
> it would just provide information on what the worker was doing at the time when the timeout triggers. So, it really only changes what information is available for logging, but not the behavior.

Maybe using the request ID would be more appropriate. Ideally, each log entry would contain the request ID as well as the worker PID, so that we can easily connect the two (*what was the worker doing* and *which request caused it to break/time out*)

service-runner.js
```diff
@@ -100,6 +102,7 @@ ServiceRunner.prototype._sanitizeConfig = function(conf, options) {
         // use the number of CPUs
         conf.num_workers = os.cpus().length;
     }
+    conf.timeout = conf.timeout || 1000;
```
d00rman: IMHO, *timeout* is too generic a parameter name. Perhaps *heartbeat_timeout* or something similar would be more appropriate.

service-runner.js
```diff
@@ -143,6 +146,24 @@ ServiceRunner.prototype.updateConfig = function updateConfig(conf) {
...
+                var lastBeat = self.workerHeartbeatTime[worker.process.pid];
+                if (!lastBeat || (!lastBeat.killed && new Date() - lastBeat.time > self.config.timeout)) {
+                    self._logger.log('error/service-runner/master',
+                        'worker ' + worker.process.pid + ' stopped sending heartbeats, killing.');
+                    self._stopWorker(workerId);
+                    // Don't need to respawn a worker, it will be restarted upon 'exit' event
+                }
+            });
+        }
+    }, self.config.timeout);
```
d00rman: I'm wondering if we should set an interval shorter than the time-out in order not to miss some workers that should be killed; e.g. in cases where this function runs slightly before a worker's time runs out, we'd have to wait twice as much to kill it. So, perhaps a skewed interval would be better here, such as `self.config.timeout / 2 + 1` (or even `/ 3 + 1`).

service-runner.js
```diff
@@ -215,25 +256,71 @@ ServiceRunner.prototype._rollingRestart = function() {
...
+                        time: new Date(),
+                        killed: false
+                    };
+                }
             });
         });
     }
 };
 
+ServiceRunner.prototype._runHeartBeat = function() {
```
d00rman: Minor nit: `_workerHeartBeat` or something similar would be a more appropriate name, so that it sends a clear message that this is executed inside the worker.

service-runner.js
Pchelolo: @d00rman I'm mostly concerned about 'false positive' killing, so now workers send a heartbeat at `timeout/3` rate. We would never miss a worker to be killed in a `timeout` interval, but what you are suggesting basically makes a timeout 2 (or even 3) times shorter than in the spec

d00rman: Hm, not really. We should still check if the last beat arrived more than `timeout` milliseconds ago. It's just that we are checking it more often.

The important thing here is that if/when the master and/or worker is super-busy, sending or checking the beat message might get scheduled way after the interval we have set. If that happens in the worker, we should obviously kill it, but if that happens in the master, it shouldn't start killing off workers because of it.

Pchelolo: @d00rman if the master is `dying busy`, the whole service is in trouble. As I understand, the whole point of node cluster lib is to avoid a situation when master is super-busy, so I think it's ok to drop this possibility from consideration as we can't do anything with it

service-runner.js
gwicke: I concur. Something making it clear that it's about workers could also be useful, like `worker_heartbeat_timeout`.

Pchelolo:
@gwicke @d00rman Updated the PR according to your comments

service-runner.js
```diff
@@ -178,19 +204,35 @@ ServiceRunner.prototype._runMaster = function() {
     // Set up rolling restarts
     process.on('SIGHUP', this._rollingRestart.bind(this));
 
-    return this._startWorkers(this.config.num_workers);
+    return this._startWorkers(this.config.num_workers)
+    .then(function(workers) {
+        setTimeout(function() {
```
d00rman: Why do you set a timer function here when `self._checkHeartBeat()` does that?

service-runner.js
```diff
@@ -178,19 +204,35 @@ ServiceRunner.prototype._runMaster = function() {
...
+        return workers;
+    });
 };
 
-ServiceRunner.prototype._stopWorker = function(id) {
-    var worker = cluster.workers[id];
+ServiceRunner.prototype._stopWorker = function(workerId) {
+    var self = this;
+    var worker = cluster.workers[workerId];
+    self.workerHeartbeatTime[worker.process.pid] = {
```
d00rman: Instead of just setting the value, you should check whether it still exists and if it's `killed` value is set to true before. If these conditions are not satisfied, you should bail out as it means the master is already waiting on the worker to die.

service-runner.js
Pchelolo: `self._checkHeartBeat()` sets an interval of 'timeout/2' which could me not enough to make completely sure all workers have enough time to start and send their first heartbeats. Here we add an additional delay (without it I've sometimes seen unnecessary restarts on startup)

service-runner.js
Pchelolo: This is checked in here before we call this func: https://github.com/wikimedia/service-runner/pull/44/files#diff-b3efe45b2faf7882e08bbdaec207491dR156

service-runner.js
```diff
@@ -215,25 +257,71 @@ ServiceRunner.prototype._rollingRestart = function() {
...
         });
     }
 };
 
+ServiceRunner.prototype._workerHeartBeat = function() {
+    // We send heart beat 3 times more frequently than check it
+    // to avoid possibility of wrong restarts
+    this.interval = setInterval(function() {
+        process.send({ type: 'heartbeat' });
+    }, this.config.worker_heartbeat_timeout / 5);
```
d00rman: 3 or 4 times suffices, IMHO. 5 seems a bit too much (also, don't forget to update the comment when you change the number :P)

service-runner.js
d00rman: Duh. True :)

service-runner.js
d00rman: Honestly, this feels like a nasty (and random) work-around. How about starting `self._checkHeartBeat()` as soon as the master is up, but then adding worker keys only once they've sent the `startup_finished` message?

Pchelolo: @d00rman Unfortunately we don't control when the worker keys are added, as we are iterating over node's internal cluster.workers collections to check each worker. Also, if we add a worker to checks only when it sends a `startup_finished`, we may never start checking a worker, which died even before sending that message. I'll try to make it start sending beats a bit earlier in the startup chain, that should help.

d00rman: The worker could send a heartbeat right before starting the loading of the service modules, couldn't it?

service-runner.js
```diff
@@ -143,6 +146,25 @@ ServiceRunner.prototype.updateConfig = function updateConfig(conf) {
...
 };
 
+ServiceRunner.prototype._checkHeartbeat = function() {
+    var self = this;
+    self.interval = setInterval(function() {
+        if (!self._shuttingDown && !self._inRollingRestart) {
+            Object.keys(cluster.workers).forEach(function(workerId) {
+                var worker = cluster.workers[workerId];
+                var lastBeat = self.workerHeartbeatTime[worker.process.pid];
+                if (!lastBeat || (!lastBeat.killed && new Date() - lastBeat.time
```
d00rman: Minor nit: `var now = new Date();` could be put before the `forEach` loop and then used here. No need to use `new Date()` for each iteration, given that all of the code is executed in the same tick anyway.

service-runner.js
```diff
@@ -153,11 +175,12 @@ ServiceRunner.prototype._runMaster = function() {
         if (!self._shuttingDown && !self._inRollingRestart) {
             var exitCode = worker.process.exitCode;
             self._logger.log('error/service-runner/master',
-                    'worker' + worker.process.pid
-                    + 'died (' + exitCode + '), restarting.');
+                    'worker ' + worker.process.pid
```
d00rman: I'm sooo happy you noticed the missing spaces surrounding the PID. They were driving me nuts a bit :))))

service-runner.js
```diff
@@ -215,25 +255,71 @@ ServiceRunner.prototype._rollingRestart = function() {
...
         var worker = cluster.fork();
         return new P(function(resolve) {
-            worker.on('message', function() {
-                resolve(self._startWorkers(--remainingWorkers));
+            fixCloseDisconnectListeners(worker);
+            worker.on('message', function(msg) {
+                if (msg.type === 'startup_finished') {
+                    resolve(self._startWorkers(--remainingWorkers));
+                } else if (msg.type === 'heartbeat') {
+                    self.workerHeartbeatTime[worker.process.pid] = {
```
d00rman: While it doesn't effectively change much, here you should check for `killed === true` for algo completeness and correctness. Imagine the master decides to kill a worker. It sets `killed = true`, but before the disconnect the worker manages to send another beat. While the worker would still be killed, its state in `self.workerHeartbeatTime` would not be reflected as such.

service-runner.js
```diff
@@ -215,25 +256,71 @@ ServiceRunner.prototype._rollingRestart = function() {
...
+            disconnect();
+        }
+        exit(exitCode, signalCode);
+    });
+}
+
 // Fork off one worker at a time, once the previous worker has finished
 // startup.
-ServiceRunner.prototype._startWorkers = function(remainingWorkers, msg) {
+ServiceRunner.prototype._startWorkers = function(remainingWorkers) {
```
d00rman: Maybe it would make sense to set up a worker's `workerHeartbeatTime` right away when forking it so that we are sure the first heartbeat happens in a timely fashion?

Pchelolo: @d00rman Not sure I understand what would be improved because of this?

d00rman: `_checkHeartbeat()` currently kills the worker right away if it doesn't have an associated `workerHeartbeatTime` entry to it, which is the correct behaviour. However, in the unlikely event that a beat check is performed between forking a worker and it having sent the first beat, it's going to be (unnecessarily) killed. Or am I missing something?

Pchelolo: Hm, indeed that's possible if we are restarting some worker. Will update

service-runner.js
```diff
@@ -215,25 +256,82 @@ ServiceRunner.prototype._rollingRestart = function() {
...
+}
+
+/**
+ * Saves the timestamp of a worker heartbeat
+ * @private
+ */
+ServiceRunner.prototype._saveBeat = function(worker) {
+    var self = this;
+    var currentVal = self.workerHeartbeatTime[worker.process.pid];
+    self.workerHeartbeatTime[worker.process.pid] = {
```
d00rman: What's the point of updating the time if the worker's `killed` flag is set to true? IMHO, we should avoid allocating a new object in that case, i.e.

```js
if (currentVal && currentVal.killed) {
    return;
}
```

config.yaml
```diff
@@ -3,6 +3,10 @@
 # Use 'ncpu' to run as many workers as there are CPU units
 num_workers: ncpu
 
+# Number of milliseconds to wait for a heartbeat from worker before killing
+# and restarting it
+worker_heartbeat_timeout: 75000
```
d00rman: Hm, that's 75 seconds, not 7.5 ;)

service-runner.js
```diff
@@ -100,6 +102,7 @@ ServiceRunner.prototype._sanitizeConfig = function(conf, options) {
         // use the number of CPUs
         conf.num_workers = os.cpus().length;
     }
+    conf.worker_heartbeat_timeout = conf.worker_heartbeat_timeout || 1000;
```
d00rman: The new default value is supposed to be 7500, not 1000

d00rman:
Published on npmjs

PR #45, opened 2015-08-27
Make service-runner more Unix-daemon-friendly

Pchelolo:
The config part is strait forward. With the logger part - we don't need to close console or gelf streams, and for file streams bunyan actually exposes a node's Writable Stream object, so it appeared to be easy to close the fd.

Bug: https://phabricator.wikimedia.org/T109727

### DIFF
--- a/service-runner.js
+++ b/service-runner.js
@@ -52,7 +52,118 @@ function ServiceRunner(options) {
 
 ServiceRunner.prototype.run = function run(conf) {
     var self = this;
-    return this.updateConfig(conf)
+    var action = P.resolve();
+    if (cluster.isMaster) {
+        action = action.then(function() {
+            return self.updateConfig(conf);
+        });
+    } else {
+        action = new Promise(function(resolve, reject) {
+            var timeout = setTimeout(function() {
+                reject(new Error('Timeout waiting for config in worker ' + process.pid));
+            }, 3000);
+            process.on('message', function(message) {
+                if (message.type === 'config') {
+                    clearTimeout(timeout);
+                    self.updateConfig(message.body)
+                    .then(resolve);
+                } else {
+                    reject(new Error('Invalid message received: ' + JSON.stringify(message)));
+                }
+            });
+        });
+    }
+    action = action
+    .then(function() {
+        if (cluster.isMaster && self.config.num_workers > 0) {
+            return self._runMaster();
+        } else {
+            return self._runWorker();
+        }
+    });
+    return action;
+};
+
+ServiceRunner.prototype._sanitizeConfig = function(conf, options) {
+    // TODO: Perform proper validation!
+    if (!conf.logging) { conf.logging = {}; }
+    if (!conf.metrics) { conf.metrics = {}; }
+    // check the number of workers to run
+    if (options.num_workers !== -1) {
+        // the number of workers has been supplied
+        // on the command line, so honour that
+        conf.num_workers = options.num_workers;
+    } else if (conf.num_workers === 'ncpu' || typeof conf.num_workers !== 'number') {
+        // use the number of CPUs
+        conf.num_workers = os.cpus().length;
+    }
+    conf.worker_heartbeat_timeout = conf.worker_heartbeat_timeout || 7500;
+    return conf;
+};
+
+
+/**
+ * Loads the config from file, serialized input or Object
+ *
+ * @param conf a configuration.
+ *             If undefined - config is loaded from config.yaml file
+ *             If Object - treated as already parsed configuration
+ *             If sting - treated a serialized yaml config
+ */
+ServiceRunner.prototype.loadConfig = function loadConfig(conf) {
+    var self = this;
+    var action;
+    if (conf && conf instanceof Object) {
+        self.config = this._sanitizeConfig(conf, self.options);
+        return P.resolve(conf);
+    } else if (conf && typeof conf === 'string') {
+        action = P.resolve(conf);
+    } else {
+        var configFile = this.options.configFile;
+        if (/^\./.test(configFile)) {
+            // resolve relative paths
+            configFile = path.resolve(self._basePath + '/' + configFile);
+        }
+        action = fs.readFileAsync(configFile);
+    }
+
+    var package_json = {};
+    try {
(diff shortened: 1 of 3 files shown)


service-runner.js
```diff
@@ -52,40 +52,36 @@ function ServiceRunner(options) {
 
 ServiceRunner.prototype.run = function run(conf) {
     var self = this;
-    return this.updateConfig(conf)
+    var action = P.resolve();
+    if (cluster.isMaster) {
+        action = action.then(function() {
```
d00rman: Simply:

```js
action = self.updateConfig(conf);
```

service-runner.js
```diff
@@ -105,44 +101,96 @@ ServiceRunner.prototype._sanitizeConfig = function(conf, options) {
...
+        if (self.options.displayVersion) {
+            console.log(name + ' ' + config.package.version);
+            process.exit(0);
+        }
+
+        // do we need to use Docker instead of starting normally ?
+        if (self.options.useDocker) {
+            self.options.basePath = self._basePath;
+            return docker(self.options, self.config);
+        }
```
d00rman: This and the previous block do not really belong here, the name of the method is `updateConfig`, but here other actions are taken. Need to move them out.

service-runner.js
```diff
@@ -105,44 +101,96 @@ ServiceRunner.prototype._sanitizeConfig = function(conf, options) {
...
+
+        // Set up the logger
+        if (!config.logging.name) {
+            config.logging.name = name;
+        }
+        self._logger = new Logger(config.logging);
```
d00rman: Idem. Additionally, this doesn't play well with receiving `SIGHUP` (cf comment there about that)

service-runner.js
```diff
@@ -202,7 +250,15 @@ ServiceRunner.prototype._runMaster = function() {
     process.on('SIGTERM', shutdown_master);
 
     // Set up rolling restarts
-    process.on('SIGHUP', this._rollingRestart.bind(this));
+    process.on('SIGHUP', function() {
+        return self.updateConfig()
+        .then(function() {
+            // Recreate loggers
+            self._logger.close();
```
d00rman: `updateConfig` already sets up a new `logger` object, so this becomes useless really. Should be good once creating a new `logger` object is moved out of `updateConfig`

Pchelolo:
@d00rman @gwicke updated

service-runner.js
```diff
@@ -105,44 +100,103 @@ ServiceRunner.prototype._sanitizeConfig = function(conf, options) {
...
+ * @param conf a configuration.
+ *             If undefined - config is loaded from config.yaml file
+ *             If Object - treated as already parsed configuration
+ *             If sting - treated a serialized yaml config
+ */
+ServiceRunner.prototype.loadConfig = function loadConfig(conf) {
     var self = this;
-    if (conf) {
+    var action;
+    if (conf && conf instanceof Object) {
```
d00rman: The two lines should be removed and just:

```js
action = P.resolve(conf);
```

Basically, the same as the next block.

service-runner.js
```diff
@@ -105,44 +100,103 @@ ServiceRunner.prototype._sanitizeConfig = function(conf, options) {
...
         self.config = this._sanitizeConfig(conf, self.options);
         return P.resolve(conf);
+    } else if (conf && typeof conf === 'string') {
+        action = P.resolve(conf);
```
d00rman: If the configuration is a string, when/where is it parsed?

d00rman: Ah, no, below. Disregard this comment, my bad.

service-runner.js
d00rman: Hm, no, indeed this case is a bit tricky, since it doesn't need parsing, but we'd still want to load the package info we add below.

service-runner.js
```diff
@@ -105,44 +100,103 @@ ServiceRunner.prototype._sanitizeConfig = function(conf, options) {
...
+            config.metrics.name = name;
+        }
+    });
+};
+
+ServiceRunner.prototype.processConfig = function() {
+    var self = this;
+    var config = self.config;
+
+    var name = config.package && config.package.name || 'service-runner';
```
d00rman: Why? We already have this in `processConfig()`

service-runner.js
```diff
@@ -202,7 +256,15 @@ ServiceRunner.prototype._runMaster = function() {
     process.on('SIGTERM', shutdown_master);
 
     // Set up rolling restarts
-    process.on('SIGHUP', this._rollingRestart.bind(this));
+    process.on('SIGHUP', function() {
```
d00rman: `SIGHUP` shouldn't be handled at all if `self.options.useDocker === true`

service-runner.js
Pchelolo: @d00rman we need this on line 188, and recalculating it is a bit more clear than use logger.name or metrics.name

service-runner.js
Pchelolo: @d00rman this is already handled by separating updateConfig and processConfig

service-runner.js
d00rman: Hm, well technically L188 might be incorrect now that it is not guaranteed `config.package` exists. We should ensure that and then remove this line, as we can be sure a name is set and use that instead (cf [this comment](https://github.com/wikimedia/service-runner/pull/45/files#r38859931))

service-runner.js
Pchelolo: But it was set before the fix, wasn't it?

d00rman: Hm, actually this PR breaks the workflow when `self.options.useDocker === true` completely. The check is done in `processConfig()` on L75, but then the code path proceeds to execute `runMaster()` regardless of this. Previously, the promise chain was broken.

Pchelolo:
@d00rman Updated so that it should work correctly. (unfortunately I can't check it, docker on mac is pretty different from normal docker, and service-runner doesn't work with it for me (neither before nor after the fix). I don't think we need to fix is, as nobody will ever deploy this on OS X.

d00rman:
Very nice work @Pchelolo !

d00rman:
Published on npmjs as v0.2.3

PR #49, opened 2015-09-28
Added a debugging log stream for pretty console logging

Pchelolo:
In [this ticket](https://phabricator.wikimedia.org/T113207) an option for pretty human-readable console logs for service-runner is requested. I find reading JSON logs painful too, so here's a `debug` stream that outputs nice color-coded logs with nice expanded stack-traces etc. 

Usage: simply add a `type: debug` stream to logger config.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -65,7 +65,7 @@ var streamConverter = {
         return {
             stream: prettyStream,
             level: stream.level || conf.level
-        }
+        };
     },
     stderr: function(stream, conf) {
         return {
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -4,7 +4,6 @@ var extend = require('extend');
 var bunyan = require('bunyan');
 var gelfStream = require('gelf-stream');
 var syslogStream = require('bunyan-syslog-udp');
-var PrettyStream = require('bunyan-prettystream');
 
 var DEF_LEVEL = 'warn';
 var LEVELS = ['trace', 'debug', 'info', 'warn', 'error', 'fatal'];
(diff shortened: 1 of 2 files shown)


d00rman:
Hm. Not sure we should use [node-bunyan-prettystream](https://github.com/mrrama/node-bunyan-prettystream), as it looks like it's not maintained any more - last commit was 2 years ago and there are 4 open issues and 11 PRs opened against it. Other forks, like [hadfieldn's](https://github.com/hadfieldn/node-bunyan-prettystream) include some of the PRs and some other improvements. The main drawback I see in the original module is the [lack of support for circular references](https://github.com/mrrama/node-bunyan-prettystream/pull/16).

d00rman:
I'm also not in favour of this being a dependency for `service-runner`. If we do choose to include a package, it should definitely go in the `devDependencies` and should be `require()`d *iff* the `debug` log type is chosen. On that note, I can't escape the feeling the name `debug` is misleading. Perhaps `pretty-out` or something similar would be more appropriate.

d00rman:
Also, note that `npm start | node_modules/.bin/bunyan` pretty-prints the logs, so I wonder if this should be even implemented.

Pchelolo:
@d00rman hm, the `npm start | node_modules/.bin/bunyan` feels like a better approach. I didn't know it can do this.

If yurik would agree on the ticket, I'll close this PR.

Pchelolo:
@d00rman Actually, I've found a use-case for this. When running tests, `npm test | node_modules/.bin/bunyan` looses mocha's color-coding and stuff, while it's still handy to see logs in pretty format in case a test is failing. So, adding this `pretty-out` mode to test configs would be a nice thing. What do you think?

d00rman:
@Pchelolo I think that's a minor corner case and that in that context we'd be just adding bloat to the package. @wikimedia/services thoughts?

gwicke:
I'm also leaning towards keeping things simple. The mocha use case seems to be pretty borderline.

Pchelolo:
I've reopened this pull request as @nyurik found one more good use-case for this capability: https://phabricator.wikimedia.org/T133611

d00rman:
Published as [v1.2.2](https://github.com/wikimedia/service-runner/tree/v1.2.2)

PR #50, opened 2015-09-28
Separated master and worker code

Pchelolo:
This is a giant PR, but it's not much happening actually. 3 new classes were added:
 - `BaseService` - abstract base class, containing common code
 - `Master` - inherits from BaseService, contains only code that runs on master. 
 - `Worker` - inherits from BaseService, contains only code that runs on worker.

`ServiceRunner` is now just a wrapper for an actual implementation (`Master` or `Worker`). The only public (exported) method is `run`.

Note: this PR only moves the code around and doesn't add any new features/fixes, so it doesn't deserve it's own release version.
Bug: https://phabricator.wikimedia.org/T113949


lib/master.js
```diff
@@ -0,0 +1,225 @@
...
+    this._shuttingDown = false;
+    // Are we performing a rolling restart?
+    this._inRollingRestart = false;
+    this.workerHeartbeatTime = {};
+}
+
+util.inherits(Master, BaseService);
+
+Master.prototype._getConfigUpdateAction = function(conf) {
+    return this._updateConfig(conf)
```
d00rman: Nit: missing semi-colon.

Pchelolo:
@d00rman May be we should enable jshint + jscs in service-runner repo, and enable travis?

d00rman:
@Pchelolo yeah, we probably should :)

PR #51, opened 2015-09-29
Enabled jshint and jscs for the repository, enable travis

Pchelolo:
jshint and jscs checks were added for the repo, this the same config as used in RESTBase.


package.json
```diff
@@ -8,9 +8,7 @@
   },
   "scripts": {
     "start": "./service-runner.js",
-    "test": "mocha",
-    "coverage": "istanbul cover _mocha -- -R spec",
-    "coveralls": "cat ./coverage/lcov.info | coveralls"
+    "test": "mocha ./test/index.js"
```
d00rman: Just `mocha` is enough.

.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: node_js
+node_js:
```
d00rman: I don't think there's a need to check for both node versions, we are only statically checking the code.

.travis.yml
```diff
@@ -0,0 +1,11 @@
...
+    - "0.12"
```
d00rman: Any reason for leaving 0.12? (Not feeling too strong about this, just being curious)

d00rman:
Needs a rebase :P

Pchelolo:
@d00rman rebased

PR #52, opened 2015-09-29
Use bunyan-syslog-udp instead of bunyan-syslog

d00rman:


d00rman:
cc @wikimedia/services

lib/logger.js
```diff
@@ -55,7 +55,6 @@ var streamConverter = {
             level: stream.level || conf.level,
             type: 'raw',
             stream: syslog_stream.createBunyanStream({
-                type: 'sys',
                 facility: syslog_stream.local0,
```
Pchelolo: We don't expose these as enum any more, so this should be `'local0'`

d00rman: Good catch!

Pchelolo:
Important note: cluster+udp [is broken in node 0.12.7](https://github.com/nodejs/node-v0.x-archive/issues/9261), so syslog logging wouldn't work (and actually the whole app would crash). There's a workaround for that, but we could avoid introducing it, because 0.12.8 is coming pretty soon and we are not using 0.12 anywhere.

Pchelolo:
LGTM

d00rman:
Yeah, I think 0.12.7 being currently broken for this specific use-case is acceptable, especially since 0.12.8 will contain [the fix for it](https://github.com/nodejs/node-v0.x-archive/pull/8643).

PR #54, opened 2015-10-02
Use yaml.dump, and bump version to v0.2.8

gwicke:
The yaml.dump would explode on an undefined value, which caused the aqs
service to fail to start. Yaml.dump lets through undefined values normally, so
startup will succeed.


d00rman:
Hm, it's been already published. We need to have an agreement on when the pkg is published and who is doing it.

gwicke:
@d00rman, we talked about it on IRC. In any case, in aqs the issue has now moved to decoding the config in client.js. Something else seems to be going on.

PR #56, opened 2015-10-07
Bug fix: Adjust test for _basePath

d00rman:
Before the master/worker split, __dirname was pointing to node_modules/service-runner, but with the split, the test is actually happening in node_modules/service-runner/lib, so adjust the test and path resolution accordingly.

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -30,8 +30,8 @@ function BaseService(options) {
     this._metrics = null;
 
     // Figure out the base path
-    this._basePath = /\/node_modules\/service-runner$/.test(__dirname) ?
-        path.resolve(__dirname + '/../../') : path.resolve('./');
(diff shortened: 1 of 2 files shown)


d00rman:
cc @wikimedia/services

lib/base_service.js
```diff
@@ -30,8 +30,8 @@ function BaseService(options) {
     this._metrics = null;
 
     // Figure out the base path
-    this._basePath = /\/node_modules\/service-runner$/.test(__dirname) ?
-        path.resolve(__dirname + '/../../') : path.resolve('./');
+    this._basePath = /\/node_modules\/service-runner/lib$/.test(__dirname) ?
```
eevans: You have an unescaped `/` here.

d00rman:
Published as v0.2.10

PR #58, opened 2015-10-16
Added support for simple expressions in num_workers config

Pchelolo:
A couple of small improvements: 
1. If `ncpu` is supplied from the console - parse and use it.
2. Added support for simple expressions like `ncpu / 2`. Eval should be safe here, as an input is checked with a restrictive regex, and also the input is provided from a config, so it's controlled by us.

Bug: https://phabricator.wikimedia.org/T108888

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -159,8 +159,8 @@ BaseService.prototype._sanitizeConfig = function(conf, options) {
         conf.num_workers = options.num_workers;
     }
     if (typeof conf.num_workers !== 'number') {
-        if (/^(?:(?:ncpu)|[\s\(\)\+\-\*\/\.\d])+$/.test(conf.num_workers)) {
-            // It's safe to make a eval here, the input format is checked
+        if (/^(?:(?:ncpu[\s\)\+\-\*\/])|[\s\(\)\+\-\*\/\.\d])+(?:ncpu)?$/.test(conf.num_workers)) {
+            // It's safe to make an eval here, the input format is checked
             /* jshint evil:true */
             var num = eval(conf.num_workers.replace(/ncpu/g, os.cpus().length));
             conf.num_workers = Math.floor(num);
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -159,11 +159,11 @@ BaseService.prototype._sanitizeConfig = function(conf, options) {
         conf.num_workers = options.num_workers;
     }
     if (typeof conf.num_workers !== 'number') {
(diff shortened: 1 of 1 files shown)


lib/base_service.js
```diff
@@ -157,9 +157,17 @@ BaseService.prototype._sanitizeConfig = function(conf, options) {
         // the number of workers has been supplied
         // on the command line, so honour that
         conf.num_workers = options.num_workers;
-    } else if (conf.num_workers === 'ncpu' || typeof conf.num_workers !== 'number') {
-        // use the number of CPUs
-        conf.num_workers = os.cpus().length;
+    }
+    if (typeof conf.num_workers !== 'number') {
+        if (/^ncpu\s*[\*\+\-\/]\s*\d+(?:\.\d+)?$/.test(conf.num_workers)) {
```
d00rman: I think I'd be better to:
- `s/ncpu/os.cpus().length`
- check that the resulting string contains only numbers, math operators, parentheses and spaces
- `eval(Math.floor(conf.num_workers))` so as to make sure the resulting number is an integer

Pchelolo: - Why?
- We already check, that an input string is in a well-defined format with a regex, why checking again?
- You mean `Math.floor(eval(...))`?

d00rman: 1. Because it is conceivable that somebody wants an expression like `(ncpu + 1) / 2`, which the current code doesn't capture.
2. Not again, but instead of. Currently only expressions like `ncpu [+-*/] number` are allowed, which is just one type of expressions
3. Yup, but both amount to the same thing :)

lib/base_service.js
```diff
@@ -157,9 +157,17 @@ BaseService.prototype._sanitizeConfig = function(conf, options) {
         // the number of workers has been supplied
         // on the command line, so honour that
         conf.num_workers = options.num_workers;
-    } else if (conf.num_workers === 'ncpu' || typeof conf.num_workers !== 'number') {
-        // use the number of CPUs
-        conf.num_workers = os.cpus().length;
+    }
+    if (typeof conf.num_workers !== 'number') {
+        if (/^(?:(?:ncpu)|[\s\(\)\+\-\*\/\.\d])+$/.test(conf.num_workers)) {
```
d00rman: `(?:ncpu[\s\)\+\-\*\/])` would be more precise, so we don't have stuff like `ncpu(`, `ncpu.2`, etc

lib/base_service.js
```diff
@@ -157,9 +157,17 @@ BaseService.prototype._sanitizeConfig = function(conf, options) {
...
+            // It's safe to make a eval here, the input format is checked
```
d00rman: `s/a/an/`

PR #61, opened 2015-10-19
Delay exiting on failure to let the logger do its work

gwicke:
`process.exit` is synchronous, while logging is async. This means that we need
to wait until the logger has logged a failure before exiting. Currently, the
logger does not support waiting for completion, so we use a 1s delay as a
work-around.


gwicke:
See also: https://github.com/trentm/node-bunyan/issues/95

PR #62, opened 2015-10-19
Release v0.2.12

gwicke:


d00rman:
Euh? Why? Bumping minor for a simple `.delay()` addition?

d00rman:
I mean that we might want to wait for another change or two before releasing a new version.

gwicke:
@d00rman, if there is anything else in the pipeline to go out soon, then waiting is fine with me. It would be nice to get fixed startup logging out there asap, as lack of startup error reporting is a major issue for our users spinning up new services.

d00rman:
> It would be nice to get fixed startup logging out there asap, as lack of startup error reporting is a major issue for our users spinning up new services.

Precisely because of that I'd be more comfortable with releasing 0.2.12 now and then increase to 0.3.0 at a later point in time. The rationale here is that `^0.2.11` (which we use throughout all of our projects) will not pick 0.3.0 up, since 0.x versions are treated differently by [npm's semver calculator](https://docs.npmjs.com/misc/semver#caret-ranges-1-2-3-0-2-5-0-0-4).

gwicke:
@d00rman, changed the version to v0.2.12.

d00rman:
Published as [v0.2.12](https://github.com/wikimedia/service-runner/tree/v0.2.12)

PR #63, opened 2015-10-20
Docker script: Ignore errors for `git rm -r node_modules`

d00rman:
If the node_modules directory has already been removed, building the repository shouldn't fail, so it is safe to ignore this specific error.


Pchelolo:
LGTM :+1:

PR #64, opened 2015-10-20
Docker script: De-duplicate module dependencies

d00rman:
Most packages use ^ and/or ~ to designate suitable dependencies. Coupling that with npm's way of searching for packages recursively up the path means that there is great benefit in de-duplicating all of service's packages when running in the container. Most notably, this vastly simplifies and minimises the deploy repository's footprint.

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -137,7 +137,7 @@ function createDockerFile() {
         contents += 'RUN mkdir /opt/service\n' +
             'ADD . /opt/service\n' +
             'WORKDIR /opt/service\n' +
-            'RUN npm install\n';
+            'RUN npm install && npm dedupe\n';
     }
 
     if (opts.uid !== 0) {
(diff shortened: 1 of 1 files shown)


d00rman:
cc @wikimedia/services

Pchelolo:
Nice idea. LGTM except a failing `jscs` check.

d00rman:
> LGTM except a failing jscs check.

Fixed now.

eevans:
This is awesome, but 'dedupe' doesn't seem to be supported by the npm packaged for Debian.  Does this mean that we'll need to install the latest from source to, for example, deploy?

d00rman:
> This is awesome, but 'dedupe' doesn't seem to be supported by the npm packaged for Debian. Does this mean that we'll need to install the latest from source to, for example, deploy?

https://gerrit.wikimedia.org/r/#/c/247727/ was created using the deploy build script with dedupe inside of a Debian Jessie container without problems.

eevans:
I... stand corrected.  I have no idea why I wasn't finding this before; Sorry for the noise!

PR #65, opened 2015-10-20
Allow to use alternative node versions

Pchelolo:
Allow docker script to use configurable node versions for build/deploy/test.

Bug: https://phabricator.wikimedia.org/T114399

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -134,22 +134,21 @@ function createDockerFile() {
         ' && rm -rf /var/lib/apt/lists/*\n';
 
     var nodeVersion = pkg.deploy.node;
+    var npmCommand = 'npm';
     if (nodeVersion && nodeVersion !== 'system') {
         var nvmDownloadURI = 'https://raw.githubusercontent.com/creationix/nvm/v0.29.0/install.sh';
         contents += 'ENV NVM_DIR /usr/local/nvm\n';
         contents += 'RUN wget -qO- ' + nvmDownloadURI + ' | bash' +
                     ' && . $NVM_DIR/nvm.sh' +
-                    ' && nvm install ' + nodeVersion +
-                    ' && nvm use ' + nodeVersion + '\n';
-        contents += 'ENV NODE_PATH=$NVM_DIR/v' + nodeVersion + '/lib/node_modules ' +
-                    'PATH=$NVM_DIR/v' + nodeVersion + '/bin:$PATH\n';
+                    ' && nvm install ' + nodeVersion + '\n';
+        npmCommand = '. $NVM_DIR/nvm.sh && nvm use ' + nodeVersion + ' && npm';
     }
 
     if (!opts.deploy) {
         contents += 'RUN mkdir /opt/service\n' +
             'ADD . /opt/service\n' +
             'WORKDIR /opt/service\n' +
-            'RUN npm install\n';
+            'RUN ' + npmCommand + ' install\n';
     }
 
     if (opts.uid !== 0) {
@@ -161,13 +160,13 @@ function createDockerFile() {
     }
 
     if (opts.deploy) {
-        contents += 'CMD npm install --production && npm install heapdump';
+        contents += 'CMD ' + npmCommand + ' install --production && npm install heapdump';
     } else if (opts.tests) {
-        contents += 'CMD ["npm", "test"]';
+        contents += 'CMD ' + npmCommand + ' test';
     } else if (opts.coverage) {
-        contents += 'CMD ["npm", "run-script", "coverage"]';
+        contents += 'CMD ' + npmCommand + ' run-script coverage';
     } else {
-        contents += 'CMD ["npm", "start"]';
+        contents += 'CMD ' + npmCommand + ' start';
     }
 
     return fs.writeFileAsync('Dockerfile', contents);
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -148,7 +148,7 @@ function createDockerFile() {
         contents += 'RUN mkdir /opt/service\n' +
             'ADD . /opt/service\n' +
             'WORKDIR /opt/service\n' +
-            'RUN ' + npmCommand + ' install && && npm dedupe\n';
+            'RUN ' + npmCommand + ' install && npm dedupe\n';
     }
 
     if (opts.uid !== 0) {
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -97,7 +97,7 @@ function createDockerFile() {
 
     var contents = '';
     var baseImg;
-    var extraPkgs = ['nodejs', 'nodejs-legacy', 'npm', 'git'];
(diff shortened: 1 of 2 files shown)


lib/docker.js
```diff
@@ -130,14 +130,26 @@ function createDockerFile() {
...
+        contents += 'ENV NODE_PATH=$NVM_DIR/v' + nodeVersion + '/lib/node_modules ' +
+                    'PATH=$NVM_DIR/v' + nodeVersion + '/bin:$PATH\n';
+    }
+
     if (!opts.deploy) {
         contents += 'RUN mkdir /opt/service\n' +
             'ADD . /opt/service\n' +
             'WORKDIR /opt/service\n' +
-            'RUN npm install\n';
+            'RUN rm -rf node_modules; npm install\n';
```
d00rman: That should be specified in a `.dockerignore` file. All of the services based off of the template [have it](https://github.com/wikimedia/service-template-node/blob/master/.dockerignore), we should add it to RESTBase as well and leave this out.

lib/docker.js
```diff
@@ -130,14 +130,26 @@ function createDockerFile() {
     }
 
     contents = 'FROM ' + baseImg + '\n' +
-        'RUN apt-get update && apt-get install -y ' + extraPkgs.join(' ') +
+        'RUN apt-get update --fix-missing && apt-get install -y ' + extraPkgs.join(' ') +
```
d00rman: Why? This shouldn't be needed, all of our container images start from a clean distro image.

Pchelolo: Hm, when I was running it locally for the first time (without cache), I've got errors on this.

Pchelolo: @d00rman retested, something was wrong back then, indeed no need for this.

lib/docker.js
```diff
@@ -141,21 +153,21 @@ function createDockerFile() {
     }
 
     if (opts.uid !== 0) {
-        contents += 'RUN groupadd -g ' + opts.gid + ' -r rungroup && ' +
-        'useradd -m -r -g rungroup -u ' + opts.uid + ' runuser\n' +
+        contents += 'RUN groupadd -r rungroup && ' +
```
d00rman: This should not go into this PR. Moreover, the UID/GID must match the local user's, which is especially important for the deploy procedure. Please remove the changes of this and the line below it.

Pchelolo: Ups, sorry, that's a workaround for testing on my mac, I forgot I didn't remove it when made a second commit.

lib/docker.js
```diff
@@ -149,13 +161,13 @@ function createDockerFile() {
     }
 
     if (opts.deploy) {
-        contents += 'CMD /usr/bin/npm install --production && /usr/bin/npm install heapdump';
+        contents += 'CMD npm install --production && npm install heapdump';
```
d00rman: this needs to be:

```js
contents += 'CMD . $NVM_DIR/nvm.sh && nvm use ' + ndoeVersion + ' && npm install --production && npm install heapdump';
```

Otherwise, the system `node` binary is used.

lib/docker.js
```diff
@@ -149,13 +161,13 @@ function createDockerFile() {
...
     } else if (opts.tests) {
-        contents += 'CMD ["/usr/bin/npm", "test"]';
+        contents += 'CMD ["npm", "test"]';
```
d00rman: Idem, turn this into:

```js
contents += 'CMD . $NVM_DIR/nvm.sh && nvm use ' + nodeVersion + ' && npm test';
```

lib/docker.js
```diff
@@ -149,13 +161,13 @@ function createDockerFile() {
...
     } else if (opts.coverage) {
-        contents += 'CMD ["/usr/bin/npm", "run-script", "coverage"]';
+        contents += 'CMD ["npm", "run-script", "coverage"]';
```
d00rman:
```js
contents += 'CMD . $NVM_DIR/nvm.sh && nvm use ' + nodeVersion + ' && npm run-script coverage';
```

lib/docker.js
```diff
@@ -149,13 +161,13 @@ function createDockerFile() {
...
     } else {
-        contents += 'CMD ["/usr/bin/npm", "start"]';
+        contents += 'CMD ["npm", "start"]';
```
d00rman:
```js
contents += 'CMD . $NVM_DIR/nvm.sh && nvm use ' + nodeVersion + ' && npm start';
```

lib/docker.js
```diff
@@ -133,6 +133,18 @@ function createDockerFile() {
...
         ' && rm -rf /var/lib/apt/lists/*\n';
 
+    var nodeVersion = pkg.deploy.node;
+    if (nodeVersion && nodeVersion !== 'system') {
+        var nvmDownloadURI = 'https://raw.githubusercontent.com/creationix/nvm/v0.29.0/install.sh';
+        contents += 'ENV NVM_DIR /usr/local/nvm\n';
+        contents += 'RUN wget -qO- ' + nvmDownloadURI + ' | bash' +
+                    ' && . $NVM_DIR/nvm.sh' +
+                    ' && nvm install ' + nodeVersion +
+                    ' && nvm use ' + nodeVersion + '\n';
```
d00rman: This is not needed, as each CMD starts a new shell / process, so this is lost.

lib/docker.js
```diff
@@ -133,6 +133,18 @@ function createDockerFile() {
...
+        contents += 'ENV NODE_PATH=$NVM_DIR/v' + nodeVersion + '/lib/node_modules ' +
```
d00rman: Not needed, see comments below.

lib/docker.js
```diff
@@ -133,6 +133,18 @@ function createDockerFile() {
...
+                    'PATH=$NVM_DIR/v' + nodeVersion + '/bin:$PATH\n';
```
d00rman: Idem.

lib/docker.js
d00rman: Actually, you need to check the value of `nodeVersion` before setting this and set it iff `nodeVersion && nodeVersion !== 'system'`

lib/docker.js
```diff
@@ -133,11 +133,22 @@ function createDockerFile() {
...
+                    ' && nvm install ' + nodeVersion + '\n';
+        npmCommand = '. $NVM_DIR/nvm.sh && nvm use ' + nodeVersion + ' && npm';
+    }
+
     if (!opts.deploy) {
         contents += 'RUN mkdir /opt/service\n' +
             'ADD . /opt/service\n' +
             'WORKDIR /opt/service\n' +
-            'RUN npm install && npm dedupe\n';
+            'RUN ' + npmCommand + ' install && && npm dedupe\n';
```
d00rman: Double `&&`

d00rman:
Very very nice work @Pchelolo !

d00rman:
Published as [v0.3.0](https://github.com/wikimedia/service-runner/tree/v0.3.0)

PR #66, opened 2015-10-28
Better error handling on promised git

Pchelolo:
Some issues with error handling on process spawn:
- `P.reject()` takes one argument, not two. And better if the argument is an `Error` instance, then the stack trace is more valid
- Notification about error in spawned process should contain `stderr`, not `stdout`
- On git spawn, we need error handlers, or the error becomes unhanded, which's reported by bluebird.

Bug: https://phabricator.wikimedia.org/T116873

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -8,7 +8,7 @@ var yaml = require('js-yaml');
 var P = require('bluebird');
 var fs = P.promisifyAll(require('fs'));
 var path = require('path');
-
+var util = require('util');
 
 // Info from the package definition
 var pkg;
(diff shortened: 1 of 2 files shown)


lib/docker.js
```diff
@@ -74,12 +87,14 @@ function promisedSpawn(args, options) {
...
-        promise = promise.catch(function(code, ret) {
+        promise = promise.catch(function(err) {
             if (options.errMessage) {
                 console.error('ERROR: ' + options.errMessage.split("\n").join("\nERROR: "));
             }
-            console.error('ERROR: ' + args.slice(0, 2).join(' ') + ' exited with code ' + code);
-            process.exit(code);
+            console.error('ERROR: ' + args.slice(0, 2).join(' ')
+                + ' exited with code ' + err.code
+                + ' and message ' + err.message);
```
d00rman: The message may be empty in some cases (user-induced errors), so this needs a guard.

PR #68, opened 2015-12-11
Logging: Support special handling for 'interesting' events

Pchelolo:
To support sample logging of 'interesting' requests, we can now specify a map of event names (or components) with the probability for them to be logged disregards the target logging level. We can't use a default bunyan logger for that purpose, so we create a special child logger for each of the components and use it.

Ticket: https://phabricator.wikimedia.org/T121163

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ logging:
   # Map keys correspond to the event names, specified in the log levels
   # For example, in 'info/webrequest', the event name is 'webrequest'.
   # Map values specify the probability for such events to be logged
-  # disregards the configured logging level.
+  # regardless of the configured logging level.
   log_components:
     webrequest: 0.2
   streams:
--- a/README.md
+++ b/README.md
@@ -87,6 +87,13 @@ worker_heartbeat_timeout: 7500
 # Logger info
 logging:
   level: info
+  # Sets up sample logging for some 'interesting' events.
+  # Map keys correspond to the event names, specified in the log levels
+  # For example, in 'info/webrequest', the event name is 'webrequest'.
+  # Map values specify the probability for such events to be logged
+  # disregards the configured logging level.
+  log_components:
+    webrequest: 0.2
   streams:
   # Use gelf-stream -> logstash
   - type: gelf
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -23,7 +23,7 @@ function Logger(confOrLogger, args) {
         self._levelMatcher = this._levelToMatcher(conf.level);
 
         // For each specially logged component we need to create
-        // a child logger that accepts everything disregards the level
+        // a child logger that accepts everything regardless of the level
         self._componentLoggers = {};
         Object.keys(self._log_components).forEach(function(component) {
             self._componentLoggers[component] = self._logger.child({
@@ -177,7 +177,7 @@ Logger.prototype._levelToMatcher = function _levelToMatcher(level) {
  * @private
  */
 Logger.prototype._getComponentLogConfig = function(level) {
-    if (level && this._log_components) {
+    if (level && Object.keys(this._log_components).length) {
         var slashIndex = level.indexOf('/');
         if (slashIndex > 0) {
             var component = level.substr(slashIndex + 1);
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -13,17 +13,32 @@ var DEF_LEVEL_INDEX = LEVELS.indexOf(DEF_LEVEL);
 
 // Simple bunyan logger wrapper
 function Logger(confOrLogger, args) {
+    var self = this;
     if (confOrLogger.constructor !== Logger) {
         // Create a new root logger
         var conf = this._processConf(confOrLogger);
-        this._logger = bunyan.createLogger(conf);
-        this._levelMatcher = this._levelToMatcher(conf.level);
+        self._log_components = conf.log_components || {};
+        delete conf.log_components;
+        self._logger = bunyan.createLogger(conf);
+        self._levelMatcher = this._levelToMatcher(conf.level);
+
+        // For each specially logged component we need to create
+        // a child logger that accepts everything disregards the level
+        self._componentLoggers = {};
+        Object.keys(self._log_components).forEach(function(component) {
+            self._componentLoggers[component] = self._logger.child({
+                component: component,
+                level: bunyan.TRACE
+            });
+        });
 
         // Set up handlers for uncaught extensions
-        this._setupRootHandlers();
+        self._setupRootHandlers();
     } else {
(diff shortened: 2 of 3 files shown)


Pchelolo:
cc @wikimedia/services

lib/logger.js
```diff
@@ -134,40 +149,96 @@ Logger.prototype._levelToMatcher = function _levelToMatcher(level) {
...
+ *
+ * If the message should be logged, returns an object with bunyan log level
+ * and a specialized logger for the given component, otherwise returns undefined
+ *
+ * @param {String} level a logging level
+ * @returns {Object|undefined} corresponding bunyan log level and a specialized logger
+ * @private
+ */
+Logger.prototype._getComponentLogConfig = function(level) {
+    if (level && this._log_components) {
```
d00rman: `this._log_components` will always evaluate to `true` (cf. line 20), so either drop it or use `Object.keys(this._log_components).length`

README.md
```diff
@@ -87,6 +87,13 @@ worker_heartbeat_timeout: 7500
 # Logger info
 logging:
   level: info
+  # Sets up sample logging for some 'interesting' events.
+  # Map keys correspond to the event names, specified in the log levels
+  # For example, in 'info/webrequest', the event name is 'webrequest'.
+  # Map values specify the probability for such events to be logged
+  # disregards the configured logging level.
```
d00rman: `s/disregards/regardless of/`

lib/logger.js
```diff
@@ -13,17 +13,32 @@ var DEF_LEVEL_INDEX = LEVELS.indexOf(DEF_LEVEL);
...
         var conf = this._processConf(confOrLogger);
-        this._logger = bunyan.createLogger(conf);
-        this._levelMatcher = this._levelToMatcher(conf.level);
+        self._log_components = conf.log_components || {};
+        delete conf.log_components;
+        self._logger = bunyan.createLogger(conf);
+        self._levelMatcher = this._levelToMatcher(conf.level);
+
+        // For each specially logged component we need to create
+        // a child logger that accepts everything disregards the level
```
d00rman: `s/disregards/regardless of/`

d00rman:
Minor nits and comments, LGTM otherwise. Needs a rebase, though.

d00rman:
Published as [v0.3.6](https://github.com/wikimedia/service-runner/tree/v0.3.6).

lib/logger.js
```diff
@@ -168,34 +231,45 @@ Logger.prototype._createMessage = function(info) {
...
+        // Got an object.
+        //
+        // We don't want to use bunyan's default error handling, as that
+        // drops most attributes on the floor. Instead, make sure we have
+        // a msg, and pass that separately to bunyan.
+        var msg = this._createMessage(info);
+
+        // Inject the detailed levelpath.
+        // 'level' is already used for the numeric level.
+        info.levelPath = level;
```
gwicke: This drops the remainder of the path.

lib/logger.js
```diff
@@ -168,34 +231,45 @@ Logger.prototype._createMessage = function(info) {
...
+};
+
+Logger.prototype.log = function(level, info) {
+    if (!info) {
+        return;
+    }
+
+    var simpleLevel = this._getSimpleLogLevel(level);
+    if (simpleLevel) {
+        this._log(info, simpleLevel, this._logger);
```
gwicke: The full levelPath also needs to be passed to `this._log`.

lib/logger.js
```diff
@@ -150,6 +167,52 @@ Logger.prototype._levelToMatcher = function _levelToMatcher(level) {
...
+ * If the message should be logged, returns an object with bunyan log level
+ * and a specialized logger for the given component, otherwise returns undefined
+ *
+ * @param {String} level a logging level
+ * @returns {Object|undefined} corresponding bunyan log level and a specialized logger
+ * @private
+ */
+Logger.prototype._getComponentLogConfig = function(level) {
+    if (level && Object.keys(this._log_components).length) {
+        var slashIndex = level.indexOf('/');
```
gwicke: This will crash if the level is not a string.

lib/logger.js
```diff
@@ -150,6 +167,52 @@ Logger.prototype._levelToMatcher = function _levelToMatcher(level) {
...
+ *
+ * If the message should be logged, returns an object with bunyan log level
+ * and a specialized logger for the given component, otherwise returns undefined
+ *
+ * @param {String} level a logging level
+ * @returns {Object|undefined} corresponding bunyan log level and a specialized logger
+ * @private
+ */
+Logger.prototype._getComponentLogConfig = function(level) {
+    if (level && Object.keys(this._log_components).length) {
```
gwicke: It would be more efficient / simpler to just set up a probability of logging a specific levelPath, and look that up in an object. This is going to be called for *a lot* of trace / debug etc events that would otherwise not be logged, so making the decision on whether to probabilistically log fast and robust is important.

lib/logger.js
```diff
@@ -13,17 +13,34 @@ var DEF_LEVEL_INDEX = LEVELS.indexOf(DEF_LEVEL);
...
+        delete conf.log_components;
+        self._logger = bunyan.createLogger(conf);
+        self._levelMatcher = this._levelToMatcher(conf.level);
+
+        // For each specially logged component we need to create
+        // a child logger that accepts everything regardless of the level
+        self._componentLoggers = {};
+        Object.keys(self._log_components).forEach(function(component) {
+            self._log_components[component] =
+                Math.ceil(self._log_components[component] * 1024);
```
gwicke: Why scale this by 1024 / use integer math?

lib/logger.js
```diff
@@ -13,17 +13,34 @@ var DEF_LEVEL_INDEX = LEVELS.indexOf(DEF_LEVEL);
...
 // Simple bunyan logger wrapper
 function Logger(confOrLogger, args) {
+    var self = this;
     if (confOrLogger.constructor !== Logger) {
         // Create a new root logger
         var conf = this._processConf(confOrLogger);
-        this._logger = bunyan.createLogger(conf);
-        this._levelMatcher = this._levelToMatcher(conf.level);
+        self._log_components = conf.log_components || {};
+        delete conf.log_components;
```
gwicke: Umm, lets not modify the config object.

Pchelolo: The original config is not modified, as it's copied wishing `_processConf` method.

lib/logger.js
Pchelolo: @gwicke According to @d00rman we will have better precision with this. The log probabilities could be extremely low with high request rates. But now I'm starting to thing this is not actually needed, since we're just comparing, all floating point issues would not be applicable here. Will remove.

PR #69, opened 2015-12-15
Logging: Improved unhandled rejection logging

Pchelolo:
A number of issue were identified in logging code due to the [mysterious EMERGENCY log messages](https://phabricator.wikimedia.org/T121296).
- Starting from bluebird 2.7.0 the recommended way to detect unhandled rejections is to [set up a global handler](http://bluebirdjs.com/docs/api/error-management-configuration.html#global-rejection-events). There's a recommendation [here](http://bluebirdjs.com/docs/api/promise.onpossiblyunhandledrejection.html)
- Bunyan processing for `Error` objects is quite different from normal objects. Although it's nice as it pretty-prints the stack and has other features, it drops all the additional fields we set in our errors. That's fixed with a custom extension for the error serialiser.

There ill be a follow up in RESTBase to avoid unhandled exceptions completely.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -94,6 +94,22 @@ Logger.prototype._processConf = function(conf) {
         conf.streams = streams;
         conf.level = LEVELS[minLevelIdx];
     }
+
+    // Define custom log message serializers
+    conf.serializers = {
+        err: function(err) {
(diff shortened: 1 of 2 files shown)


Pchelolo:
cc @wikimedia/services

PR #70, opened 2015-12-16
Worker status logging on worker death

Pchelolo:
It's generally useful to be able to log a worker death reason together with the death notification message. In particular, this is one of the missing nits that we need to be able to move parsoid to service-runner. 

This patch adds a capability for a service to emit `service_status` messages, that would be tracked, and on worker death the last one would be logged as possible reason of the death. It's possible to enhance that, by, for example, tracking which of the services sent which message, or by reporting N last statuses, but for now the initial implementation should be fine I suppose.

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -39,6 +39,23 @@ Worker.prototype._getConfigUpdateAction = function() {
                 reject(new Error('Invalid message received: ' + JSON.stringify(message)));
             }
         });
+
+        if (cluster.isWorker) {
+            // If got a status update in a worker - forward it to the master
+            process.on('service_status', function(message) {
+                try {
+                    process.send({
+                        type: 'service_status',
+                        status: message
+                    });
+                } catch (e) {
+                    self._logger.log('warn/service-runner/worker/', {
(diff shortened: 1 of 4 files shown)


Pchelolo:
cc @wikimedia/services

lib/worker.js
```diff
@@ -39,6 +39,23 @@ Worker.prototype._getConfigUpdateAction = function() {
...
+
+        if (cluster.isWorker) {
+            // If got a status update in a worker - forward it to the master
+            process.on('service_status', function(message) {
+                try {
+                    process.send({
+                        type: 'service_status',
+                        status: message
+                    });
+                } catch (e) {
```
Pchelolo: We need a `try-catch` here as the circular references in the message will throw an error.

gwicke: Makes sense.

Pchelolo:
Published as service-runner@0.3.4

d00rman:
Please remember to also tag the release and push it on upstream/master when publishing. Did so for this one [here](https://github.com/wikimedia/service-runner/tree/v0.3.4).

PR #71, opened 2015-12-17
Batch requests to metrics

Pchelolo:
`statsd` supports batching, but the `node-statsd` client we use doesn't support that, so I've made a PR to support batch mode in the client. I'm going to post the PR upstream, but that repo repo didn't have many attention, so we could use a custom fork for now.

Here is the PR with `node-statsd` patch required for this to work: https://github.com/Pchelolo/node-statsd/pull/1 (it's not a PR to upstream, just across my own branches. Created to simplify the review of the changes). 

I've tested this making a full wikipedia-like stats setup (with a proxy), running restbase benchmarks and tests in both batching and non-batching modes and comparing the results. Everything looks fine.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -47,7 +47,8 @@ function makeStatsD(options, logger) {
         txstatsd: options.type === 'txstatsd',
         globalize: false,
         cacheDns: true,
-        mock: false
+        mock: false,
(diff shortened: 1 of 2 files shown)


Pchelolo:
cc @wikimedia/services

PR #72, opened 2015-12-18
Logging: Improved message if no message was provided

Pchelolo:
Before the patch the logging message for logged objects would be `[object Object]`, which is not very nice. Now we set a message to `Message not supplied` which's cleaner.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -155,6 +155,19 @@ Logger.prototype.child = function(args) {
     return new Logger(this, newArgs);
 };
 
+Logger.prototype._createMessage = function(info) {
+    var msg = info.msg || info.message || info.info;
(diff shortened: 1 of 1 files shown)


lib/logger.js
```diff
@@ -155,6 +155,19 @@ Logger.prototype.child = function(args) {
     return new Logger(this, newArgs);
 };
 
+Logger.prototype._createMessage = function(info) {
+    var msg = info.msg || info.message || info.info;
+    if (msg) {
+        return msg;
+    }
+    var infoStr = '' + info;
```
d00rman: Now that we have a separate method for this, let's be explicit and say:

```js
var infoStr = info.toString();
```

PR #73, opened 2016-01-13
Use babeljs for backwards compatibility on nodejs < 4

gwicke:
We are starting to use ES2015 features available in the current node 4 LTS
release, but would still like to be backwards-compatible with node 0.10 and
0.12 if that is reasonably possible. This patch wires up babeljs to bring JS
on par with node 4's features using the es2015-node4 preset. It only enables
babel if node is older than v4.x, so does not affect performance if node >= 4
is used.

While performance on 0.10 seems generally okay so far, baseline memory usage
is significantly higher with babel (about 3x), likely due to the use of the
require hook and in-memory caching. According to the documentation it sounds
like pre-compiling the app & then running it from disk would reduce memory
usage, but since this is more invasive I am leaving that optimization for
later.


Pchelolo:
LGTM

eevans:
> ... baseline memory usage is significantly higher with babel (about 3x)...

3x, Yikes!

d00rman:
@gwicke, rebase so we can get this in before publishing v1.0.0?

gwicke:
@d00rman, I would vote for waiting with this until after 1.0.0. The slow-down is quite substantial, and I'm not convinced that it would actually be very useful to anybody in practice in its current state.

arlolra:
Maybe close this after #103

gwicke:
@arlolra, agreed that this is not going to happen in service-runner. If we go with something like https://github.com/gwicke/dist-es6/tree/multi_target, then it'll be used to compile service modules in a pre-publish hook. Closing.

PR #74, opened 2016-01-14
Metrics: Support metrics batching using the hot-shots client

Pchelolo:
Switch from `node-statsd` client, which doesn't support batching and is not maintained to [hot-shots](https://www.npmjs.com/package/hot-shots) client, which is fork that supports batching and seem to be actively developed. 

Also, exposed public batching options interface.

cc @wikimedia/services

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,9 +102,12 @@ logging:
 
 # Statsd metrics reporter
 metrics:
-  type: txstatsd
+  type: statsd
   host: localhost
   port: 8125
+  batch: # Metrics batching options. Supported only for `statsd` reporter type
+    max_size: 1500 # Max size of the batch buffer (default: 1500)
+    max_delay: 1000  # Max delay for an individual metric in milliseconds (default: 1000)
 
 services:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -95,9 +95,12 @@ logging:
 
 # Statsd metrics reporter
 metrics:
-  type: txstatsd
+  type: statsd 
   host: localhost
   port: 8125
+  batch: # Metrics batching options. Supported only for `statsd` reporter type
+    max_size: 10000 # Max size of the batch buffer (default: 1500)
```
gwicke: Should we recommend jumbo packets? Why not stay with the safe-ish 1.5k default?

Pchelolo: @gwicke updated

README.md
```diff
@@ -95,9 +95,12 @@ logging:
 
 # Statsd metrics reporter
 metrics:
-  type: txstatsd
+  type: statsd
```
d00rman: Extra space at the end?

PR #75, opened 2016-01-19
Logging: Fix issues with log_components feature

Pchelolo:
There were some issues with https://github.com/wikimedia/service-runner/pull/68 , so making a follow up.
Issues fixed:
- Full levelPath is logged now
- Don't do the integer math on probabilities as it doesn't seem to be needed
- Fixed the code which makes the decision whether to log or not - more robust now.

Also, I've noticed a problem, that when the logged message is a string, default logger args are not logged to it - also fixed.

cc @wikimedia/services

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -21,13 +21,12 @@ function Logger(confOrLogger, args) {
         delete conf.log_components;
         self._logger = bunyan.createLogger(conf);
         self._levelMatcher = this._levelToMatcher(conf.level);
+        self._allLevelsMatcher = this._levelToMatcher('trace');
 
         // For each specially logged component we need to create
         // a child logger that accepts everything regardless of the level
         self._componentLoggers = {};
         Object.keys(self._log_components).forEach(function(component) {
-            self._log_components[component] =
-                Math.ceil(self._log_components[component] * 1024);
             self._componentLoggers[component] = self._logger.child({
                 component: component,
                 level: bunyan.TRACE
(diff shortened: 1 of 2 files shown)


lib/logger.js
```diff
@@ -174,22 +175,19 @@ Logger.prototype._levelToMatcher = function _levelToMatcher(level) {
...
-            if (logProbability && Math.ceil(Math.random() * 1024) < logProbability) {
-                return {
-                    level: level.substring(0, slashIndex),
-                    logger: this._componentLoggers[component]
-                };
-            }
+Logger.prototype._getComponentLogConfig = function(levelPath) {
+    var logProbability = this._log_components[levelPath];
+    if (logProbability && Math.random() < logProbability) {
+        var levelMatch = this._allLevelsMatcher.exec(levelPath);
```
d00rman: Is this check really needed? `_allLevelsMatcher` is set to `trace`, so this will always be true.

lib/logger.js
```diff
@@ -232,14 +230,14 @@ Logger.prototype._createMessage = function(info) {
...
-Logger.prototype._log = function(info, level, logger) {
+Logger.prototype._log = function(info, level, levelPath, logger) {
     if (info instanceof String) {
         // Need to convert to primitive
         info = info.toString();
     }
 
     if (typeof info === 'string') {
-        logger[level].call(logger, info);
+        logger[level].call(logger, extend({ msg: info }, this.args));
```
d00rman: Nice!

lib/logger.js
Pchelolo: It's needed to extract the simple `level` from the `levelPath`, for example `warn` from `warn/parsoid/blablabla`

d00rman: `levelPath.split('/').shift()` ?

d00rman: `levelPath.split('/')[0]` is even faster :)

Pchelolo: Surprisingly, the regex is 2 times faster that a split solution.

Pchelolo: @d00rman updated

PR #78, opened 2016-01-26
Drop txstatsd support

gwicke:
We don't use txstatsd any more, and aren't aware of any third party users
either. Lets simplify matters by reducing choices.


Pchelolo:
:+1:

PR #79, opened 2016-01-26
Add the ability to create prefixed child metric clients

gwicke:
We would like to be able to prefix all metrics emitted as part of a request
with a fixed prefix, so that we can distinguish internal / external or update
requests & all sub-requests associated with it.

Similar to the logger instance, this patch adds a `makeChild` method on
returned metric clients, which given a name will create such a prefixed metric
client.

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -37,7 +37,7 @@ function LogStatsD(logger, serviceName) {
 // Minimal StatsD wrapper
 function makeStatsD(options, logger) {
     var statsd;
-    var srvName = normalizeName(options.name);
+    var srvName = options._prefix ? options._prefix : normalizeName(options.name);
     var statsdOptions = {
         host: options.host,
         port: options.port,
@@ -61,6 +61,15 @@ function makeStatsD(options, logger) {
     } else {
         statsd = new StatsD(statsdOptions);
     }
+
+    // Support creating sub-metric clients with a fixed prefix. This is useful
+    // for systematically categorizing metrics per-request, by using a
+    // specific logger.
+    statsd.makeChild = function(name) {
+        var childOptions = Object.assign({}, options);
+        childOptions._prefix = statsdOptions.prefix + normalizeName(name) + '.';
+        return makeStatsD(childOptions, logger);
+    };
 
     // Add a static utility method for stat name normalization
     statsd.normalizeName = normalizeName;
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -67,7 +67,7 @@ function makeStatsD(options, logger) {
     // specific logger.
     statsd.makeChild = function(name) {
         var childOptions = Object.assign({}, options);
-        childOptions._prefix = statsdOptions.prefix + normalizeName(name) + '.';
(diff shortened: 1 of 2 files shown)


Pchelolo:
:+1:

lib/statsd.js
```diff
@@ -62,6 +62,15 @@ function makeStatsD(options, logger) {
         statsd = new StatsD(statsdOptions);
     }
 
+    // Support creating sub-metric clients with a fixed prefix. This is useful
+    // for systematically categorizing metrics per-request, by using a
+    // specific logger.
+    statsd.makeChild = function(name) {
+        var childOptions = Object.assign({}, options);
+        childOptions._prefix = statsdOptions.prefix + normalizeName(name) + '.';
```
d00rman: Should probably check that `statsdOptions.prefix` evaluates to a truth-y value.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "service-runner",
-  "version": "0.3.9",
+  "version": "0.4.0",
```
d00rman: This version isn't incompatible with previous ones, so maybe release it as 0.3.10 instead?

gwicke: It changes the public interface in a backwards-compatible manner, so according to semver the minor version ought to be incremented.

lib/statsd.js
gwicke: It's guaranteed to be a string that passed through normalizeName, so there shouldn't be a need to check this here.

package.json
gwicke: Although, http://semver.org/#spec-item-4 basically says "anything goes" if major is < 1. 

I think we should move to 1.x soon to reflect that service-runner is now providing some reasonable stability guarantees.

gwicke:
Updated to 1.0.0. Lets hope this is a winner.

lib/statsd.js
d00rman: I don't see where this is checked. But, to be more exact, if the prefix is null / undefined / empty string, then the name of the parent and child is equal, which should be a bug, IMHO, since as it stands, one does not have to declare the prefix options in order to use `makeChild()`.

Also, the parent metrics object's name does not end with a dot, while the child's name will (as presented on line 40).

gwicke: The prefix is set at https://github.com/gwicke/service-runner/blob/makeChild/lib/statsd.js#L40, which will only work if it's a string.

d00rman: Only `options.name` is ever normalised, `options._prefix` isn't; it's only used as-is. Also, what's up with the dot (cf. my previous comment) ? :P

gwicke: options._prefix is a private attribute (as indicated by `_`) that we construct from normalized parts.

Regarding the dot, there was indeed an issue in that we ended up with two trailing dots for child metrics clients. Removed the trailing dot from the _prefix construction to avoid this. Good catch!

d00rman:
Published as [v1.0.0](https://github.com/wikimedia/service-runner/tree/v1.0.0) :+1: :+1: :+1:

gwicke:
@d00rman, thanks! Lets document & announce 1.0 soon.

PR #80, opened 2016-02-02
Expose config directory to the service

Pchelolo:
See https://github.com/wikimedia/restbase/pull/495

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,14 @@ var ServiceRunner = require('service-runner');
 new ServiceRunner().run();
 ```
 
+All file paths in the config are relative to the application base path. Base path
+is an absolute path to the folder where your application is located (where `package.json` file is located).
+
+By default, we assume that your project depends on `service-runner` and you follow standrad
+node project layout. However, if custom layout is used, you could override the app base path with:
+- `APP_BASE_PATH` environment variable
+- `app_base_path` config stanza.
+
 We are also working on a [standard
 template](https://github.com/wikimedia/service-template-node) for node
 services, which will set up this & other things for you.
--- a/README.md
+++ b/README.md
@@ -66,11 +66,13 @@ var ServiceRunner = require('service-runner');
 new ServiceRunner().run();
 ```
 
-All file paths in the config are relative to the application base path. Base path
-is an absolute path to the folder where your application is located (where `package.json` file is located).
+All file paths in the config are relative to the application base path. 
+The base path is an absolute path to the folder where your application 
+is located (where `package.json` file is located).
 
-By default, we assume that your project depends on `service-runner` and you follow standrad
-node project layout. However, if custom layout is used, you could override the app base path with:
+By default, we assume that your project depends on `service-runner` and 
+you follow standard node project layout. However, if a custom layout is 
+used, you must override the app base path with either:
 - `APP_BASE_PATH` environment variable
 - `app_base_path` config stanza.
 
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -28,11 +28,19 @@ function BaseService(options) {
 
     this._logger = null;
     this._metrics = null;
-
-    // Figure out the base path
-    this._basePath = /\/node_modules\/service-runner\/lib$/.test(__dirname) ?
-        path.resolve(__dirname + '/../../../') : path.resolve('./');
 }
+
+BaseService.prototype._getAppBasePath = function(config) {
+    if (process.env.APP_BASE_PATH) {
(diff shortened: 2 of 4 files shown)


d00rman:
I'm OK with this change, but we still have the fatal flaw in the code base where we only *guess* what the base path actually is, but cannot really know it. How about passing it in either as a config option or env var (even better, perhaps) ?

Pchelolo:
@d00rman We were not using env vars before (with one exception of `RB_TEST_BACKEND`, which is a test-only thing), so I'm not sure whether we should make an exception about it, so I think adding a `base_path` config property is more in line with other stuff. I'm not sure though whether we should require it, or default to guessing..

d00rman:
To be honest, I'd include all of the three options:
- if the env var is provided, use it
- if the config stanza is present, use it
- otherwise fall back to *best-guessing*

Makes sense?

Pchelolo:
@d00rman Yep, if we clearly document that. 

`Entering bike-shedding territory`
Does `APP_BASE_PATH` as an env var and `app_base_path` as a config stanza sound good to you?
`Leaving bike-shedding territory`

lib/worker.js
```diff
@@ -106,7 +106,7 @@ Worker.prototype._run = function() {
         var modName = service.module || service.name;
         if (/^\./.test(modName)) {
             // resolve relative paths
-            modName = path.resolve(self._basePath + '/' + modName);
+            modName = path.resolve((self.config.config_dir || self._basePath), modName);
```
Pchelolo: So... In normal operation mode the config file location will be known. However, tests do some magic on config files, then require a `service-runner` and pass the config file directly as an object. Obviously, in this case the location of a config file is unknown, so we default back to our guessed `base_path`. 

This works, but I'm still not very happy about all the black magic happening here and open to suggestions. My be it would be better to completely prohibit passing the config as object to service-runner and in tests make config modifications and write the results into a temp file and pass it's path?

Pchelolo:
Hm.. Actually I'm wondering whether I should mark it as 1.1.0, or even 2.0, as it's a completely backwards incompatible change. Now the service main file path is relative to the config location, not the guessed `basePath`, so everything using this will be broken... Which is a bad idea actually.

lib/base_service.js
```diff
@@ -199,7 +207,7 @@ BaseService.prototype._loadConfig = function _loadConfig(conf) {
         var configFile = this.options.configFile;
         if (/^\./.test(configFile)) {
             // resolve relative paths
-            configFile = path.resolve(self._basePath + '/' + configFile);
+            configFile = path.resolve(process.cwd() + '/' + configFile);
```
Pchelolo: This is actually more precise, the relative path to the config should be resolved relative to the working directory, which is an expected behaviour.

lib/base_service.js
```diff
@@ -28,12 +28,20 @@ function BaseService(options) {
 
     this._logger = null;
     this._metrics = null;
-
-    // Figure out the base path
-    this._basePath = /\/node_modules\/service-runner\/lib$/.test(__dirname) ?
-        path.resolve(__dirname + '/../../../') : path.resolve('./');
 }
 
+BaseService.prototype._getAppBasePath = function(config) {
```
d00rman: This is more of a setter method, so I'd `s/get/set/`

d00rman:
@Pchelolo could you also add `app_base_path` to the docs and a note about how to use it?

README.md
```diff
@@ -66,6 +66,14 @@ var ServiceRunner = require('service-runner');
 new ServiceRunner().run();
 ```
 
+All file paths in the config are relative to the application base path. Base path
```
d00rman: "The base path"

README.md
```diff
@@ -66,6 +66,14 @@ var ServiceRunner = require('service-runner');
...
+is an absolute path to the folder where your application is located (where `package.json` file is located).
+
+By default, we assume that your project depends on `service-runner` and you follow standrad
```
d00rman: "standard"

README.md
```diff
@@ -66,6 +66,14 @@ var ServiceRunner = require('service-runner');
...
+node project layout. However, if custom layout is used, you could override the app base path with:
```
d00rman: "if a custom"
"you must override"

d00rman: "with either"

d00rman:
Published as [v1.1.0](https://github.com/wikimedia/service-runner/tree/v1.1.0)

gwicke:
The revenge of Express, by getting another "APP" in there ;)

PR #81, opened 2016-02-03
Metrics: Decreased the default maximum batch size

Pchelolo:
We need to decrease the default maximum batch size to account for IP and UDP headers, so that packets don't exceed MTU. In case of IPv6 it's 48 bytes, so 1450 looks safe enough (we even have 2 bytes 'reserved').

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -1,6 +1,14 @@
 'use strict';
 var StatsD = require('hot-shots');
 
+/**
+ * Maximum size of a metrics batch used by default.
+ *
+ * @const
+ * @type {number}
+ */
+var DEFAULT_MAX_BATCH_SIZE = 1450;
+
 var nameCache = {};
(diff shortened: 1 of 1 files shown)


lib/statsd.js
```diff
@@ -1,6 +1,14 @@
 'use strict';
 var StatsD = require('hot-shots');
 
+/**
+ * Maximum size of a metrics batch used by default.
+ *
+ * @const
+ * @type {number}
+ */
+var DEFAULT_MAX_BATCH_SIZE = 1450;
```
gwicke: This would be a good use case for a `const`, rather than a `var`.

Pchelolo: hehe.. Actually `var` in ES6 is considered legacy as there's not real use for it. So, when we drop node 0.10, which, I hope, will happen soon, we would probably replace `var` with `const` and `let` everywhere..

PR #82, opened 2016-02-03
Use new hot-shots package version, update other dependencies

Pchelolo:
The [hot-shots PR](https://github.com/brightcove/hot-shots/pull/14) was merged and a new version was released. So, update our dependency and release new version of service-runner.

cc @wikimedia/services

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "service-runner",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "description": "Generic nodejs service supervisor / cluster runner",
   "main": "service-runner.js",
   "bin": {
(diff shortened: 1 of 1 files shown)


gwicke:
@Pchelolo, mind adding the new version number (v1.1.1) in the commit summary line? It's often useful to be able to grep for such versions later.

gwicke:
With all those dependency changes we'll need to be careful with how we deploy this.

d00rman:
> With all those dependency changes we'll need to be careful with how we deploy this.

AFAIK, only the bluebird update can have a real impact in prod

Pchelolo:
@gwicke All the dependencies had `^` prefix, so effectively we're using those versions right now.

gwicke:
@Pchelolo, I see. If we are already running 3.2.1 in production, then no concerns from me.

Given that we had bluebird memory leaks being introduced in minor versions before, perhaps we should switch it to the `~` pattern across the board?

d00rman:
You are right @Pchelolo :

```
$ grep version /srv/deployment/restbase/deploy/node_modules/bluebird/package.json 
  "version": "3.2.1",
```

> Given that we had bluebird memory leaks being introduced in minor versions before, perhaps we should switch it to the ~ pattern across the board?

It would be wise to be more conservative with bluebird, :+1:

gwicke:
@Pchelolo, mind adjusting the pattern? Otherwise this looks ready to go to me.

Pchelolo:
@gwicke hehe, if I just the pattern here we'd need to just it everywhere else too... Which is problematic :)

gwicke:
@Pchelolo, we can probably follow up with the other packages before the next bluebird release comes out.

gwicke:
Published as v1.1.1.

PR #83, opened 2016-02-04
T125747: Use _stopWorker in master shutdown

arlolra:


lib/master.js
```diff
@@ -65,7 +65,10 @@ Master.prototype._run = function() {
         if (self.interval) {
             clearInterval(self.interval);
         }
-        cluster.disconnect(function() {
+        P.each(Object.keys(cluster.workers), function(workerId) {
```
gwicke: It might be okay to stop all at once, with P.all / map.

gwicke:
:+1: 

/cc @wikimedia/services

arlolra:
`P.map(Object.keys(cluster.workers), self._stopWorker)` would have been sufficient :(

d00rman:
Indeed @arlolra . Weird we we haven't spotted it sooner.

arlolra:
I pushed 2012443f7c214288a1c3b6bc6a8f0a1cabfe39d4.

PR #84, opened 2016-02-04
Module loading: Support relative paths without a leading dot

Pchelolo:
Currently when loading modules we detect relative paths by the leading dot. However, in other places we support relative paths with no leading dot, so I've updated the module loading code to have the same semantics as we have elsewhere. 

1. If there's a leading `/` - it's an absolute path - require.
2. Otherwise - try to require it as a module, if failed - treat it as a relative path.

cc @wikimedia/services

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -270,4 +270,19 @@ BaseService.prototype._updateConfig = function _updateConfig(conf) {
     });
 };
 
+BaseService.prototype._requireModule = function(modName) {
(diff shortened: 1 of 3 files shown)


d00rman:
Tested locally with RESTBase, not waiting for Travis.

d00rman:
Published as [v1.1.2](https://github.com/wikimedia/service-runner/tree/v1.1.2)

PR #86, opened 2016-02-06
Bind _stopWorker to self to avoid crash on master shutdown

Pchelolo:
We didn't bind `_stopWorker` to self, so `this` was undefined which led to master crash on shutdown

cc @wikimedia/services


d00rman:
Published as [v1.1.3](https://github.com/wikimedia/service-runner/tree/v1.1.3)

PR #89, opened 2016-02-24
Rate limiter support

gwicke:
Add a high-performance rate limiter backed by
https://github.com/gwicke/limitation. Backends include a local in-memory
backend, as well as a Kademlia DHT.

Task: https://phabricator.wikimedia.org/T125123

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,15 @@ metrics:
     max_size: 1500 # Max size of the batch buffer (default: 1500)
     max_delay: 1000  # Max delay for an individual metric in milliseconds (default: 1000)
 
+# Rate limiter (enabled by default)
+ratelimit:
+  type: memory
+  # optional: Kademlia backend
+  # type: kad
+  # seeds:
+  #  - 123.456.789.233
+  #  - 456.789.90.2
+
 services:
   - name: parsoid
     # a relative path or the name of an npm package, if different from name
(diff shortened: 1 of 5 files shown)


lib/ratelimiter.js
```diff
@@ -0,0 +1,120 @@
...
+    }, self._interval);
+}
+
+/**
+ * Send current counters to the master process.
+ */
+RateLimiterWorker.prototype._sendToMaster = function() {
+    var counters = this._counters;
+    this._counters = {};
+    process.send({
```
Pchelolo: Why do we wrap `worker.send` with try-catch on line 22, and do not wrap process.send here?

lib/ratelimiter.js
```diff
@@ -0,0 +1,120 @@
...
+RateLimiterWorker.prototype.isAboveLimit = function(key, limit, increment) {
+    var counter = this._counters[key];
+    if (!counter) {
+        counter = this._counters[key] = {
+            value: 0,
+            limits: {},
+        };
+    }
+    counter.value += increment || 1;
+    counter.limits[limit] = counter.limits[limit] || Date.now();
```
Pchelolo: What's the point `Date.now?`. Where is that used?

README.md
```diff
@@ -172,6 +181,52 @@ options.metrics.histogram('my_histogram', 42, ['foo', 'bar']);
...
+  # Optional
+  # Address / port to listen on
+  # Default: localhost:3050, random port fall-back if port used
+  listen:
+    address: localhost
+    port: 3050
+  # Counter update / block interval; Default: 10000ms
+  interval: 10000
+```
+
```
Pchelolo: Could be nice to also add an API reference, e.g. describe the `isAboveLimit` method.

Pchelolo:
LGTM overall,just a couple of nits and misunderstandings.

lib/ratelimiter.js
gwicke: It's possible for a worker to die between enumerating workers & sending the message, and a failing send should not take down the master in that case. However, if the master dies, there isn't much left to do in a worker. It will just exit.

lib/ratelimiter.js
gwicke: It is used to remove watches for old limits when processing blocks.

README.md
gwicke: It is described at https://github.com/wikimedia/service-runner/pull/89/files#diff-04c6e90faac2675aa89e2176d2eec7d8R193.

PR #90, opened 2016-03-04
Invert logic to match method name, and release v1.1.6

gwicke:


Pchelolo:
So, no we're defaulting to allowing access right? 

If that's the case - LGTM

gwicke:
This is just reflecting the rename from check() to isAboveLimit(). While both `limitation` and `service-runner` had been updated to the new name, only `limitation` had the return value flipped in line with the name change.

Pchelolo:
:+1:

PR #92, opened 2016-03-17
Attach volume for tests

Pchelolo:
When using `docker-test` command, some services could have a need to create files in the service dir. An example is `restbase` which is tested in SQLite mode in docker and so needs to create a db file. Without attaching the volume it fails to do so, so the tests are failing.

cc @wikimedia/services

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -211,8 +211,8 @@ function buildImg() {
 /**
  * Starts the container and returns once it has finished executing
  *
- * @param {Array} args the array of extra parameters to pass, optional
- * @param {Boolean} hidePorts whether to keep the ports hidden inside the container, optional
+ * @param {Array} [args] the array of extra parameters to pass, optional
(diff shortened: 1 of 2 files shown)


lib/docker.js
```diff
@@ -527,6 +527,9 @@ function main(options, configuration) {
     .then(function() {
         if (opts.deploy) {
             return updateDeploy();
+        } else if (opts.tests) {
+            // append a volume so that tests could create files in container
+            return startContainer(['-v', process.cwd() + ':/opt/service']);
```
d00rman: When running tests, we already use `/opt/service` as the root dir for the service, where the node modules are reinstalled from scratch. Simply attaching the `cwd` over it means that incompatibilities may arise if the host and guest OSes are different, as binary modules might not be compatible. A better solution might be to `chown` and `chmod` that directory so that tests may create files as needed.

PR #94, opened 2016-04-12
Enable metrics batching by default

Pchelolo:
Metrics batching was enabled for restbase for a long period of time - enable it by default for other services.

cc @wikimedia/services


d00rman:
Published as [v1.2.1](https://github.com/wikimedia/service-runner/tree/v1.2.1)

gwicke:
:+1:

PR #96, opened 2016-05-18
Handle worker deaths on startup properly

Pchelolo:
In case the worker fails to start, we have to continue startup sequence.

cc @wikimedia/services 

Bug: https://phabricator.wikimedia.org/T135615

### DIFF
--- a/lib/master.js
+++ b/lib/master.js
@@ -24,8 +24,7 @@ function Master(options) {
     this._shuttingDown = false;
     // Are we performing a rolling restart?
     this._inRollingRestart = false;
-    this._firstWorkerStarted = false;
-    this._firstWorkerStartupAttempts = 0;
+    this._currentStartingWorker = undefined;
     this.workerStatusMap = {};
 }
 
@@ -48,7 +47,9 @@ Master.prototype._run = function() {
         + this.config.num_workers + ' workers');
 
     cluster.on('exit', function(worker) {
-        if (!self._shuttingDown && !self._inRollingRestart) {
+        if (!self._shuttingDown
+                && !self._inRollingRestart
+                && self._currentStartingWorker !== worker.process.pid) {
             var exitCode = worker.process.exitCode;
             var info = {
                 message: 'worker ' + worker.process.pid + ' died (' + exitCode + '), restarting.'
@@ -188,6 +189,7 @@ Master.prototype._startWorkers = function(remainingWorkers) {
     var self = this;
     if (remainingWorkers) {
         var worker = cluster.fork();
+        self._currentStartingWorker = worker.process.pid;
         self._saveBeat(worker);
         return new P(function(resolve) {
             fixCloseDisconnectListeners(worker);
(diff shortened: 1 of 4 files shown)


lib/master.js
```diff
@@ -45,14 +47,14 @@ Master.prototype._run = function() {
...
-    cluster.on('exit', function(worker, code, signal) {
+    cluster.on('exit', function(worker) {
         if (!self._shuttingDown && !self._inRollingRestart) {
             var exitCode = worker.process.exitCode;
             var info = {
                 message: 'worker ' + worker.process.pid + ' died (' + exitCode + '), restarting.'
             };
             if (self.workerStatusMap[worker.process.pid]
-                    && self.workerStatusMap[worker.process.pid].status) {
+                && self.workerStatusMap[worker.process.pid].status) {
```
d00rman: Nit: let's re-indent it to make it clear this is part of the `if` clause, not the block.

lib/master.js
```diff
@@ -88,9 +90,9 @@ Master.prototype._run = function() {
         .then(self._rollingRestart.bind(self));
     });
 
-    var setupPromise = P.resolve();
+
```
d00rman: Nit: no need for this extra line

lib/master.js
```diff
@@ -195,9 +197,35 @@ Master.prototype._startWorkers = function(remainingWorkers) {
                 type: 'config',
                 body: yaml.dump(self.config)
             });
+
+            var workerExit = function(code) {
+                if (self._firstWorkerStarted) {
+                    self._logger.log('warn/service-runner/master', 'worker ' + worker.process.pid
+                        + ' died during startup (' + code + '), continue');
```
d00rman: `continue`? It's not clear what it stands for.

service-runner.js
```diff
@@ -36,5 +36,5 @@ module.exports = ServiceRunner;
 
 if (module.parent === null) {
     // Run as a script: Start up
-    return new ServiceRunner().run();
+    new ServiceRunner().run();
```
d00rman: Why removing the `return` ?

Pchelolo: new version of JSCS doesn't like it, it's not a function, so it doesn't allow it.

Pchelolo:
@d00rman Amended according to your comments

d00rman:
LGTM in principle, even though I will need to test it a bit.

Pchelolo:
@d00rman I've done some testing with randomised startup failures and it looked good. Also try to kill workers and do rolling restarts, all seemed to work OK.

lib/master.js
```diff
@@ -195,9 +195,35 @@ Master.prototype._startWorkers = function(remainingWorkers) {
                 type: 'config',
                 body: yaml.dump(self.config)
             });
+
+            var workerExit = function(code) {
```
d00rman: Hm, this procedure breaks the sequential fork of workers. Instead of letting the normal worker restart procedure handle that, wouldn't it be better simply not to decrease `remainingWorkers` ?

Pchelolo: @d00rman Then we need to move registration of the `cluster.on('exit'...` after all workers were started. In that case, if a worker dies after it was started, but before the last worker finished startup, we would loose a worker.

d00rman: Yup. I'd be :+1: on doing this. Conceptually, worker restarts per se shouldn't happen before the completion of the start-up sequence.

Pchelolo: @d00rman however, I have no idea how to achieve that..

d00rman: Let's think about it tomorrow :)

gwicke: We could actually re-evaluate the purely sequential startup strategy. The main reason for starting up sequentially was to avoid concurrent schema changes. Since only the first worker normally performs those changes, it should be okay to switch to parallel startup after the first worker is up.

d00rman: @gwicke Change prop wouldn't like that. Its start-up is already flaky as-is, giving up on its sequentiality would only increase it.

d00rman: @Pchelolo one possible idea would be to have a `this._isStartup` flag which is set to `true` only on master process start and upon receiving `SIGHUP`. If that flag is active, the `cluster.on('exit')` handler becomes a no-op. Then, once `remainingWorkers` reaches `0`, the flag is set to `false`.

Pchelolo: @d00rman This is equivalent to setting the cluster-level exit handler after all workers were started, so makes it possible to loose workers. 

Conceptually, there're 2 different exit reasons: 
- The worker failed to start during startup - this one should be handled by the `workerExit` handler
- The worker started up normally, but then died before all other workers were started. This one is handled by cluster-level exit handler. 

Unfortunately we can't separate one from another, so if we disable cluster-level handler during startup we will loose workers.

d00rman:
:+1:

d00rman:
Published as [v1.2.3](https://github.com/wikimedia/service-runner/tree/v1.2.3)

PR #97, opened 2016-05-18
Execute service-runner.js when forking

arlolra:
/cc @wikimedia/services

### DIFF
--- a/service-runner.js
+++ b/service-runner.js
@@ -19,6 +19,11 @@ var Worker = require('./lib/worker');
 // sudo sysctl -w net.core.somaxconn=4096
 cluster.schedulingPolicy = cluster.SCHED_NONE;
 
+// When forking, we should execute this script.
+if (cluster.isMaster) {
+    cluster.setupMaster({ exec: __filename });
+}
+
 function ServiceRunner(options) {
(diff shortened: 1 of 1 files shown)


arlolra:
https://phabricator.wikimedia.org/P3127

arlolra:
👎 hold off on merging while I update it

d00rman:
This looks like a no-op to me. If the master process managed to reach this code, what prevents the worker to do the same since a `fork()` replicates the call?

arlolra:
The default value for `exec` is `process.argv[1]`.  This matters when you're using `service-runner` as module, not the bin.

https://nodejs.org/api/cluster.html#cluster_cluster_setupmaster_settings

arlolra:
For example, in https://github.com/wikimedia/restbase/blob/e7fb889ca00efdf320c3b6f968077c26f3f8087f/test/features/metrics.js#L10

you wouldn't want the worker to take the same route to reaching this code.

d00rman:
Ah right, you're addressing the situation where service-runner is used programmatically but the number of workers is non-zero.

arlolra:
Yes, exactly.

PR #98, opened 2016-05-19
Fail master if it cannot start any workers

Pchelolo:
In case master cannot start any workers for 3 times - kill the process as it's clearly see buggy code.

I have been testing this for quite a while with different services, killing workers, doing restarts etc, and it seems to work well.

cc @wikimedia/services 

Bug: https://phabricator.wikimedia.org/T134002

### DIFF
--- a/lib/master.js
+++ b/lib/master.js
@@ -94,6 +94,8 @@ Master.prototype._run = function() {
 
     // Set up rolling restarts
     process.on('SIGHUP', function() {
+        self._firstWorkerStarted = false;
+        self._firstWorkerStartupAttempts = 0;
         return self._updateConfig()
         .then(function() {
             // Recreate loggers
--- a/lib/master.js
+++ b/lib/master.js
@@ -12,6 +12,16 @@ var Logger = require('./logger');
 var RateLimiterMaster = require('./ratelimiter').master;
 
 /**
+ * Maximum number of startup attempts. In case the first worker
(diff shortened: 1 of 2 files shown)


lib/master.js
```diff
@@ -199,8 +201,24 @@ Master.prototype._startWorkers = function(remainingWorkers) {
             });
 
             var workerExit = function(code) {
-                self._logger.log('warn/service-runner/master', 'worker ' + worker.process.pid
-                    + ' died during startup (' + code + '), continue startup');
+                if (!self._firstWorkerStarted && self._firstWorkerStartupAttempts++ >= 3) {
```
d00rman: Let's make the number of attempts a constant for increased visibility.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "service-runner",
-  "version": "1.2.3",
+  "version": "1.3.0",
```
d00rman: Could we bump the version in a dedicated PR? I'd like to get in a really small fix for start-up module loading which doesn't really deserve its own version :P

d00rman:
LGTM :+1: modulo a minor comment and question.

d00rman:
You also need to reset `firstWorkerStarted` and `_firstWorkerStartupAttempt` (mamma mia, the name of this variable is soooo long) when SIGHUP is received.

PR #99, opened 2016-05-19
Clarify run() return value & support service shutdown with stop()

gwicke:
- Consistently return worker startup results, with or without workers
  (`num_workers != 0`). Require results to be JSON.stringify()-able, and
  support returning a `close()` method for each service.
- Add a `stop()` method on ServiceRunner, and close each service in each
  worker (if any) when it is called. This relies on a `close()` method being
  returned from services, which currently is the case for `http.Server`
  instances as returned by hyperswitch and the service-template.

Bug: https://phabricator.wikimedia.org/T135679

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,10 +72,10 @@ Example script for **starting** a service, using commandline options:
 
 ```javascript
 var ServiceRunner = require('service-runner');
-new ServiceRunner().run();
+new ServiceRunner().start();
 ```
 It is also possible to skip commandline options, and pass in a config
-directly to `ServiceRunner.run()` (see [the config section](#config_loading)
+directly to `ServiceRunner.start()` (see [the config section](#config_loading)
 for details on available options). Here is an example demonstrating this, as
 well as return values & the `stop()` method:
 
@@ -83,7 +83,7 @@ well as return values & the `stop()` method:
 var ServiceRunner = require('service-runner');
 var runner = new ServiceRunner();
 
-var startupPromise = runner.run({
+var startupPromise = runner.start({
     num_workers: 0,
     services: [{
         name: 'parsoid',
--- a/README.md
+++ b/README.md
@@ -12,7 +12,9 @@ module.exports = function (options) {
     // Metrics reporter (statsd,log)
     var metrics = options.metrics;
 
-    // Start the app, returning a promise
+    // Start the app, returning a promise.
+    // Return an object with a `close()` function for clean shut-down support.
+    // (ex: node's HTTP server instances).
     return startApp(config, logger, metrics);
 }
 ```
@@ -48,7 +50,7 @@ Options:
 ```bash
 npm install --save service-runner
 ```
-
+### As a binary
 In package.json, configure `npm start` to call service-runner:
 ```javascript
   "scripts": {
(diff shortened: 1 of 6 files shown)


gwicke:
Tested with hyperswitch.

README.md
```diff
@@ -60,29 +62,53 @@ module parameter to your service's entry point.
...
-used, you must override the app base path with either:
-- `APP_BASE_PATH` environment variable
-- `app_base_path` config stanza.
-
-We are also working on a [standard
-template](https://github.com/wikimedia/service-template-node) for node
-services, which will set up this & other things for you.
+```javascript
+var ServiceRunner = require('service-runner');
+var runner = new ServiceRunner();
```
arlolra: I found that you really need to pass in `new ServiceRunner({ num_workers: -1 })` here, otherwise yargs tries to set options from argv, which isn't expected in a library.

gwicke: The doc changes in this patch actually describe that behavior, but you are right that we should perhaps reconsider it. Lets make that a separate task / PR?

Pchelolo:
We'd need to update change-prop tests too, they use close feature too.

lib/master.js
```diff
@@ -214,7 +225,8 @@ Master.prototype._startWorkers = function(remainingWorkers) {
                     case 'startup_finished':
                         worker.removeListener('exit', workerExit);
                         self._currentStartingWorker = undefined;
-                        resolve(self._startWorkers(--remainingWorkers));
+                        res.push(msg.serviceReturns[0]);
```
Pchelolo: Why only the first element? As I understand the format of the return value could be
```
[ [ worker1.service1, worker1.service2 ], [worker2.service1, worker2.service2] ]
```

lib/master.js
```diff
@@ -97,9 +107,9 @@ Master.prototype._run = function() {
     .then(function() {
         return self._startWorkers(self.config.num_workers);
     })
-    .then(function(workers) {
+    .then(function(workerReturns) {
```
Pchelolo: Could replace `.then` with `.tap` here.

lib/master.js
```diff
@@ -34,11 +34,26 @@ Master.prototype._getConfigUpdateAction = function(conf) {
     return this._updateConfig(conf);
 };
 
+Master.prototype.stop = function() {
+    var self = this;
+    if (self.config.num_workers === 0) {
+        // No workers needed, run worker code directly.
+        return Worker.prototype.stop.call(self);
```
Pchelolo: May be wrap it to array, for consistency with num_workers > 0 case?

d00rman:
This is not compatible with the tests used by the service template or any of its derivatives, as they expect [to receive the `http` object of the service back](https://github.com/wikimedia/service-template-node/blob/dceac08c7d1b3dd8aef2d84427de0fe8c02a980b/test/utils/server.js#L54-L62).

Given that this is a clear change of service-runner's interface, IMO such a change can be introduced only by bumping the major version, i.e. releasing this as v2.0.0. This would allow us to avoid breakage with most services (note that `1.3.0 ~ ^1.x.x` in semver).

lib/master.js
gwicke: The worker returns an array of arrays too, for consistency in the `num_workers = 0` case. When we combine those in a single array, we need to unwrap the single element as a result.

README.md
arlolra: The doc change says `ServiceRunner.run()` though, and I'm talking about the constructor.

But yeah, not a blocker here / separate issue.  I was just sayin'

gwicke:
@d00rman, changed the version to 2.0.0.

arlolra:
While you're bumping the major version number, can you consider getting rid of core-js?

gwicke:
@arlolra, do you have a specific reason for using an old version of core-js, or are you planning on not using it any more at all? Have you seen specific issues with current core-js?

arlolra:
I just don't think service-runner should be altering the execution environment.  It may be ok (apart from the caveats below) as a binary, but using service-runner in the library form yields really unexpected behaviour.

* service-runner isn't the only entrypoint into our codebase (think parse.js) so we need to include core-js anyways, but now we can't update it independent of service-runner (what are the consequences of require two different version of it?)
* I don't yet know if there are any specific issue, it needs its own round of auditing, testing, rt run and production deploy (working on that now though)

This is pretty much what I said in https://phabricator.wikimedia.org/T135176#2305130

/cc @wikimedia/parsing

d00rman:
@arlolra makes a good point. `service-runner` doesn't need the shims per se (from what I can tell, we only really use it in [cmd-line arg checking](https://github.com/wikimedia/service-runner/blob/baecdc3d1e2fe436089de9982734e89ff1a5cb45/lib/base_service.js#L131-L133) but that's a trivial case), so why alter the environment when there's no need to? Services that do make us of `core-js` can easily include it themselves.

Pchelolo:
@d00rman We also use `Object.assign` in one place, that's not available in 0.10. 

Regarding the shims - I think the step 0 is to make a decision on the lowest node version supported by service-runner, and them modify the code to use only stuff supported on that version out of the box.

d00rman:
> @d00rman We also use Object.assign in one place, that's not available in 0.10.

That's something we can easily polyfill in the form of a utility function (without altering `Object` itself).

> Regarding the shims - I think the step 0 is to make a decision on the lowest node version supported by service-runner, and them modify the code to use only stuff supported on that version out of the box.

Given that the functionality and the feature list of `service-runner` have been pretty stable for some time now, it doesn't seem likely we'll be introducing stuff that explicitly relies on node > 0.1x, so I think that sticking with 0.1x all the while getting rid of `core-js` is acceptable.

Pchelolo:
@arlolra @d00rman I've created a separate PR for the `core-js` question. Let's not mix 2 unrelated issues together.

d00rman:
> @arlolra @d00rman I've created a separate PR for the core-js question. Let's not mix 2 unrelated issues together.

Yup, makes sense. Thnx @Pchelolo ! But we all agree on ditching core-js, that one should definitely be merged before this one, as this is a major-version bump.

Pchelolo:
@d00rman I think a major version bump could include both PRs

d00rman:
> @d00rman I think a major version bump could include both PRs

Yes, hence merge the other one first (but without any version bump), so that when the bump does happen here, it's already in, given that this PR bumps the version officially.

subbuss:
https://gerrit.wikimedia.org/r/#/c/289508/ is currently blocked on these 2 PRs being merged and a major version bump for service-runner. The new version will let us clean up some of the code in gerrit. /cc @arlolra

gwicke:
One final question before we merge this: Currently the API is `.run()` and `.stop()`. Should we rename `.run()` to `.start()`, and keep it working as a (deprecated) alias for a full major version, emitting a warning log line?

arlolra:
Sounds good to me.

gwicke:
Rebased on master. @wikimedia/services @wikimedia/parsing, this is good to go.

subbuss:
Minor: not sure if easy to fix, but in the commit message for " Rename `run()` to `start()` ", s/remode/removed/

gwicke:
@subbuss, good catch! Fixed the commit message.

gwicke:
Merged, published & tagged as v2.0.0. @wikimedia/parsing, this should un-block you.

arlolra:
Thanks @gwicke

PR #102, opened 2016-05-20
Fix package.json's licence field to hold a valid SPDX value

d00rman:


d00rman:
cc @wikimedia/services

Pchelolo:
👍

PR #103, opened 2016-05-30
Removed core-js dependency

Pchelolo:
In #99 we've been discussing the `core-js` dependency, and got to an agreement that it has to be removed from `service-runner`. This PR removes it, and shims manually all places where functions, not supported in node 0.10 are used. 

It will be a no-op for all `hyperswitch` services, since it load up `core-js` in it's index file. Although, for other services this might be an API-change, so we have to bump the version accordingly. I'm not sure whether to bump it to 2.0.0 or 1.4.0 though..

cc @wikimedia/services

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -9,6 +9,13 @@ var StatsD = require('hot-shots');
  */
 var DEFAULT_MAX_BATCH_SIZE = 1450;
 
+function objectAssign(target, source) {
+    Object.keys(source).forEach(function(keyName) {
+        target[keyName] = source[keyName];
+    });
+    return target;
+}
+
 var nameCache = {};
 function normalizeName(name) {
     // See https://github.com/etsy/statsd/issues/110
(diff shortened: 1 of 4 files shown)


subbuss:
You could include this PR as part of other big service-runner upates and bump it to 2.0.0

Pchelolo:
@subbuss Ye, I think that's the best plan. Just moved it out of the #99 as it's an unrelated change.

lib/statsd.js
```diff
@@ -9,6 +9,13 @@ var StatsD = require('hot-shots');
  */
 var DEFAULT_MAX_BATCH_SIZE = 1450;
 
+function objectAssign(target, source) {
+    Object.keys(source).forEach(function(keyName) {
```
d00rman: Perhaps it'd be worth checking if `Object.assign` exists, and if it does use that, otherwise fall back to the polyfill

gwicke:
Merged in preparation for 2.0 release.

PR #104, opened 2016-05-31
Docker: set the IN_DOCKER env var when in Docker; release v1.3.1

d00rman:
Sometimes services need to know if they are running inside Docker or
not in order to alter their environment. For example, the Mobile Content
Service uses request recording/playback, which should not be used when
running in Docker, so set the environment variable IN_DOCKER to signal
that.

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -177,6 +177,8 @@ function createDockerFile() {
         contents += 'ENV HOME=/root/ LINK=g++\n';
     }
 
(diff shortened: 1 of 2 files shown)


d00rman:
cc @wikimedia/services

d00rman:
Published as [v1.3.1](https://github.com/wikimedia/service-runner/tree/v1.3.1)

PR #106, opened 2016-06-03
Return all values when more than one service is started

arlolra:
The README says,

> // startupResults is an array of arrays of objects returned by each
> // service. These objects should be JSON.stringify()-able.

/cc @gwicke

### DIFF
--- a/lib/master.js
+++ b/lib/master.js
@@ -66,7 +66,10 @@ Master.prototype._start = function() {
     if (self.config.num_workers === 0) {
         // No workers needed, run worker code directly.
         // FIXME: Create a Master / Worker dynamically on _start()?
-        return Worker.prototype._start.call(self);
+        return Worker.prototype._start.call(self).then(function(serviceReturns) {
+            // Wrap to match _startWorkers
+            return [serviceReturns];
+        });
     }
 
     // Fork workers.
@@ -121,9 +124,8 @@ Master.prototype._start = function() {
     .then(function() {
         return self._startWorkers(self.config.num_workers);
     })
-    .then(function(workerReturns) {
+    .tap(function() {
         self._checkHeartbeat();
-        return workerReturns;
     });
 };
 
@@ -230,7 +232,7 @@ Master.prototype._startWorkers = function(remainingWorkers, res) {
                     self._logger.log('fatal/service-runner/master',
                         'startup failed, exiting master');
                     // Don't exit right away, allow logger to process message
-                    setTimeout(function() {  process.exit(1); }, 1000);
+                    setTimeout(function() { process.exit(1); }, 1000);
                     return;
                 }
 
@@ -246,7 +248,7 @@ Master.prototype._startWorkers = function(remainingWorkers, res) {
                 // Let all the exit listeners fire before reassigning current worker ID
                 process.nextTick(function() {
                     self._currentStartingWorker = undefined;
-                    resolve(self._startWorkers(remainingWorkers));
+                    resolve(self._startWorkers(remainingWorkers, res));
                 });
             };
 
@@ -257,7 +259,7 @@ Master.prototype._startWorkers = function(remainingWorkers, res) {
                         worker.removeListener('exit', workerExit);
                         self._currentStartingWorker = undefined;
                         self._firstWorkerStarted = true;
-                        res.push(msg.serviceReturns[0]);
+                        res.push(msg.serviceReturns);
                         resolve(self._startWorkers(--remainingWorkers, res));
                         break;
                     case 'heartbeat':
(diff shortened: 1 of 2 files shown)


lib/worker.js
```diff
@@ -164,7 +164,7 @@ Worker.prototype._start = function() {
         self.serviceReturns = res;
         // Make sure that only JSON-serializable values are returned.
         try {
-            return [JSON.parse(JSON.stringify(res))];
+            return JSON.parse(JSON.stringify(res));
```
arlolra: `res` is already an array here.

lib/master.js
```diff
@@ -246,7 +246,7 @@ Master.prototype._startWorkers = function(remainingWorkers, res) {
                 // Let all the exit listeners fire before reassigning current worker ID
                 process.nextTick(function() {
                     self._currentStartingWorker = undefined;
-                    resolve(self._startWorkers(remainingWorkers));
+                    resolve(self._startWorkers(remainingWorkers, res));
```
arlolra: is this right?

lib/worker.js
gwicke: Yes, but for consistency with the master it's expected to be an array of arrays. In the case of running without workers, this will be the exact return value.

lib/master.js
```diff
@@ -257,7 +257,8 @@ Master.prototype._startWorkers = function(remainingWorkers, res) {
                         worker.removeListener('exit', workerExit);
                         self._currentStartingWorker = undefined;
                         self._firstWorkerStarted = true;
-                        res.push(msg.serviceReturns[0]);
+                        res.push(msg.serviceReturns.length > 1 ?
+                            msg.serviceReturns : msg.serviceReturns[0]);
```
gwicke: This would mix return values from services in different workers, rather than keeping them grouped.

arlolra: `serviceReturns` is an array of return values from the different services started in the worker. Currently, you're dropping all but the first one on the floor.

lib/worker.js
arlolra: I think it makes more sense to return it this way, and then in master near

    if (self.config.num_workers === 0) {
        // No workers needed, run worker code directly.
        // FIXME: Create a Master / Worker dynamically on _start()?
        return Worker.prototype._start.call(self);
    }

wrap that in an array.

If you really wanted to match master as it currently is, it should be `[JSON.parse(JSON.stringify(res))[0]]` but see below.

Also, what about the error case here.  That's just `[e]`.

arlolra: Force pushed the suggestion above.

lib/master.js
arlolra: The README says,

> // startupResults is an array of arrays of objects returned by each service

so this should really just be `msg.serviceReturns`.  Do you prefer that?

arlolra: Force pushed to match the README.

lib/worker.js
gwicke: > If you really wanted to match master as it currently is, it should be [JSON.parse(JSON.stringify(res))[0]] but see below.

No, this would only return the startup return from the first service started in the worker, and not all of them. Service-runner supports starting several services in each worker, so we need to return an array of startup returns per worker.

gwicke:
I think it might be helpful to clarify the terminology here:

- With services, we tend to describe network services as listed in the config file. Service-runner can start multiple of them in a single process / worker ([example](https://github.com/wikimedia/mediawiki-node-services/blob/master/config.yaml)).
- Workers are optional cluster workers. Each worker runs the same set of services specified in the config.

This is why the return value is designed to be an array of arrays: Each inner array is expected to hold the results for all services started up in a worker / process.

lib/master.js
gwicke: I don't think so; this is inside an event handler that reacts to a worker exiting, so something that normally happens after startup. Workers are normally restarted until they successfully start up & send a return message, which will be returned after all workers have successfully started up.

lib/worker.js
arlolra: But that's exactly what this whole patch fixes, you aren't doing that because you're pushing `msg.serviceReturns[0]`

gwicke: Right, but we are getting an array of arrays in the master, as we wrapped it in the worker.

arlolra: Not when `num_workers !== 0` though.

lib/master.js
arlolra: If any workers succeed though (and thus decrement remaining workers) you'll lose their return values.

gwicke: On reflection, I think you actually have a good point here. Looks good to me.

PR #107, opened 2016-06-24
Support switching on trace logging for 5 seconds

Pchelolo:
We need this for change-prop life debugging.

cc @wikimedia/services

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -9,6 +9,9 @@ var DEF_LEVEL = 'warn';
 var LEVELS = ['trace', 'debug', 'info', 'warn', 'error', 'fatal'];
 var DEF_LEVEL_INDEX = LEVELS.indexOf(DEF_LEVEL);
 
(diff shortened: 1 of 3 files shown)


lib/logger.js
```diff
@@ -268,11 +275,20 @@ Logger.prototype._log = function(info, level, levelPath, logger) {
...
     if (!level || !info) {
         return;
     }
 
-    var simpleLevel = this._getSimpleLogLevel(level);
+    if (Logger.logTrace) {
+        simpleLevel = extractSimpleLevel(level);
+        if (simpleLevel) {
+            this._log(info, simpleLevel, level, this._traceLogger);
+            return;
```
d00rman: Should be out of the `if` block, since if `extractSimpleLevel` can't get the level, `_getSimpleLogLevel` won't be able to do so either.

d00rman:
Published as [v2.0.2](https://github.com/wikimedia/service-runner/tree/v2.0.2)

PR #108, opened 2016-07-03
Update README.md

nyurik:
so much pain to get it to work :(


d00rman:
This should not be part of the example config, but rather of the docs themselves.

gwicke:
Is there really much of a use case for `bunyan-prettystream`, considering that `bunyan` is installed by default?

Pchelolo:
@gwicke There's a couple of neat use-cases for that, see https://phabricator.wikimedia.org/T133611

PR #109, opened 2016-07-08
Log the deprecation warning only if the logger is present; v2.0.3

d00rman:
When building dokcer containers, the logger component is not
initialised. So, if the client started the build process using the
deprecated `ServiceRunner.run()` method, the process will fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "service-runner",
-  "version": "2.0.2",
+  "version": "2.0.3",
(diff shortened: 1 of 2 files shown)


d00rman:
cc @wikimedia/services

Pchelolo:
👍

d00rman:
Published as [v2.0.3](https://github.com/wikimedia/service-runner/tree/v2.0.3)

PR #110, opened 2016-07-19
Bug fix: Serialise the worker returns before sending them; rel v2.0.4

d00rman:
The worker tries to correctly serialise and deserialise the return
values before returning them. However, when there are non-serialisable
values, the worker will encounter a fatal exception when trying to send
the object to the master. This commit fixes this issue by trying to
serialise the return object first.

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -165,17 +165,19 @@ Worker.prototype._start = function() {
         });
     })
     .then(function(res) {
-        // Signal that this worker finished startup
-        if (cluster.isWorker) {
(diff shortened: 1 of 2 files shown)


d00rman:
cc @wikimedia/services

d00rman:
Self-merging as this bug precludes us from using service-runner in services returning the server object.

d00rman:
Published as [v2.0.4](https://github.com/wikimedia/service-runner/tree/v2.0.4)

PR #113, opened 2016-08-01
Support building deploy repo on Mac

Pchelolo:
Recently native docker for Mac beta was released (https://www.docker.com/products/docker#/mac) which removes the need to have a VirtualBox machine to host the docker daemon processes. Additionally, they've created `osxfs` (https://docs.docker.com/docker-for-mac/osxfs/) - the proxy FS that makes access to the host filesystem easy from the container. So now to support building the deploy repos on OSX we can simply run as root inside the container and all the permissions will be handled for us but docker/osxfs.

Bug: https://phabricator.wikimedia.org/T103490

cc @wikimedia/services

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -262,8 +262,10 @@ function ensureDockerVersion() {
             process.exit(1);
         }
 
-        if (os.type() === 'Darwin' && semver.lt(dockerVersion[1], '1.12.0')) {
-            console.error('Building the deploy repo on OSX supported only with docker 1.12.0+');
+        var minimumDockerVersion = os.type() === 'Darwin' ? '1.12.0' : '1.8.0';
+        if (semver.lt(dockerVersion[1], minimumDockerVersion)) {
+            console.error('Building the deploy repo on OSX supported only with docker '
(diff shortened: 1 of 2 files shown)


lib/docker.js
```diff
@@ -243,7 +248,26 @@ function startContainer(args, hidePorts) {
...
+            process.exit(1);
+        }
+
+        dockerVersion = /(\d+\.\d+\.\d+)/.exec(dockerVersion);
+        if (!dockerVersion[1]) {
+            console.error("Couldn't find docker version. Make sure docker is installed");
+            process.exit(1);
+        }
 
+        if (os.type() === 'Darwin' && semver.lt(dockerVersion[1], '1.12.0')) {
```
d00rman: Nice check! Could you add one for the non-Darwin case as well? We need docker >= 1.8.0

Pchelolo: @d00rman Done.

d00rman:
Published as [v2.0.6](https://github.com/wikimedia/service-runner/tree/v2.0.6)

PR #114, opened 2016-08-03
Docker: don't capture git-review output

Pchelolo:
Sometimes `git review` needs user interaction, for example if the gerrit username/password is not set up, or if for some reason it's trying to submit several changes. In that case if we capture the output for the user it looks like the script is stuck. At least with this user would be able to interact with git-review and figure out what's the problem.

Bug: https://phabricator.wikimedia.org/T141917
cc @wikimedia/services

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -249,22 +249,16 @@ function startContainer(args, hidePorts) {
 }
 
 function ensureDockerVersion() {
-    return promisedSpawn(['docker', '-v'], { capture: true })
+    return promisedSpawn(['docker', 'version', '--format', '{{.Server.Version}}'], { capture: true })
     .then(function(dockerVersion) {
         if (!dockerVersion) {
             console.error('Docker is not found.');
             process.exit(1);
         }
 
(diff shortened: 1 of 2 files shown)


d00rman:
How about we make it dependant on whether the user specified the `--verbose` cmdline arg? Without that, the context may not be clear enough.

Pchelolo:
@d00rman The context would be pretty clear, it will print. `Submitting your changes for review` and then whatever git-review output would be, so it would be either success message, or a prompt with a question to the user. 

These question asking situation indicates that something is configured incorrectly, so asking a use to provide yet another flag is imho an overkill as you would first run without a flag and see how the process gets stuck for no reason - start waiting and loose time before you understand to kill the process and restart with  `--verbose`

Pchelolo:
Introduced one more improvement in the script - docker version check was done incorrectly before on linux

d00rman:
Published as [v2.0.7](https://github.com/wikimedia/service-runner/tree/v2.0.7)

PR #115, opened 2016-08-09
Docker: reset the submodule after switching back to master

d00rman:


d00rman:
cc @wikimedia/services

Pchelolo:
Not particularly sure what issue are you trying to solve here, but it wouldn't hurt

d00rman:
@Pchelolo after submitting the patch, the script checks out the master branch, but the submodule stays on the patch rev id, so git shows as if though you had outstanding commits. Not really an issue per se, more of a small correctness facelift.

PR #116, opened 2016-08-10
Support providing env variables for the docker script

Pchelolo:
Here's why I need this: the new kafka driver by default builds `librdkafka` from sources, which is fine for normal clients / local testing etc, since you don't expect the user to install binary packages. However, for production we are going to use debian package for `librdkafka` and build the native node bindings against it. That is controlled by setting a `BUILD_LIBRDKAFKA` env variable during the build - so we need service-runner to be able to set this variable.

cc @wikimedia/services

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -185,7 +185,7 @@ function createDockerFile() {
     var envCommand = 'ENV IN_DOCKER=1';
     if (pkg.deploy.env && Object.keys(pkg.deploy.env)) {
         Object.keys(pkg.deploy.env).forEach(function(envVar) {
-            envCommand += ' ' + envVar + ' ' + pkg.deploy.env[envVar];
+            envCommand += ' ' + envVar + '=' + pkg.deploy.env[envVar];
         });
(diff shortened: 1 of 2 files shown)


lib/docker.js
```diff
@@ -182,7 +182,13 @@ function createDockerFile() {
         contents += 'ENV HOME=/root/ LINK=g++\n';
     }
 
-    contents += 'ENV IN_DOCKER=1\n';
+    var envCommand = 'ENV IN_DOCKER=1';
+    if (pkg.deploy.env && Object.keys(pkg.deploy.env)) {
+        Object.keys(pkg.deploy.env).forEach(function(envVar) {
+            envCommand += ' ' + envVar + '=' + pkg.deploy.env[envVar];
```
d00rman: Perhaps better to quote the values so as not to cause problems with string containing spaces

d00rman:
Published as [v2.0.8](https://github.com/wikimedia/service-runner/tree/v2.0.8)

PR #117, opened 2016-08-11
Docker: Do not remove .txt files

d00rman:
Some packages use .txt files as data files, so don't remove them
forcefully.


d00rman:
cc @wikimedia/services

PR #118, opened 2016-08-11
Support providing custom sources for packages

Pchelolo:
Bug: https://phabricator.wikimedia.org/T142666

Here's an example how Change-Prop deploy config would look like:
```json
"deploy": {
    "node": "4.4.6",
    "target": "debian",
    "env": {
      "BUILD_LIBRDKAFKA": "0"
    },
    "dependencies": {
      "debian": [
        {
          "repo_url": "https://apt.wikimedia.org/wikimedia",
          "release": "jessie-wikimedia",
          "pool": "main",
          "packages": [ "librdkafka-dev" ]
        }
      ]
    }
```

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -118,6 +118,11 @@ function createDockerFile() {
     var contents = '';
     var baseImg;
     var extraPkgs = ['nodejs', 'nodejs-legacy', 'npm', 'git', 'wget'];
+    // An array where we store custom-sourced extra packages
+    // Each element has 'repo_url', 'release', 'pool' and 'packages' properties
+    // 'packages' is an array of package names
+    // 'repo_url' and 'pool' combinations are unique
+    var customSourcePkgs = [];
 
     // set some defaults
     if (!pkg.deploy) {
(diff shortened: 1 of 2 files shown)


lib/docker.js
```diff
@@ -135,24 +140,50 @@ function createDockerFile() {
...
-    Array.prototype.push.apply(extraPkgs, pkg.deploy.dependencies._all);
 
-    if (!baseImg || baseImg === '') {
-        console.error('ERROR: You must specify a valid target!');
-        console.error('ERROR: Check the deploy stanza in package.json and targets.yaml');
-        process.exit(2);
+    if (customSourcePkgs.some(function(customSourcePkgSpec) {
+            return /^https/.test(customSourcePkgSpec.repo_url);
+        })) {
+        extraPkgs.push('apt-transport-https');
```
d00rman: Awesome! Kudos for this!

lib/docker.js
```diff
@@ -135,24 +140,50 @@ function createDockerFile() {
...
+        'RUN apt-get update && apt-get install -y ' + extraPkgs.join(' ') +
         ' && rm -rf /var/lib/apt/lists/*\n';
 
+    customSourcePkgs.forEach(function(customSourcePkgSpec) {
+        contents += 'RUN echo "deb ' + customSourcePkgSpec.repo_url +
+                ' ' + customSourcePkgSpec.release +
+                ' ' + customSourcePkgSpec.pool + '" > /etc/apt/sources.list;' +
+                ' apt-get update; apt-get install -y --force-yes -t ' +
+                customSourcePkgSpec.release + ' ' + customSourcePkgSpec.packages.join(' ') +
+                ' && rm -rf /var/lib/apt/lists/*\n';
```
d00rman: It's actually much more stable for APT to remove the cache only after all of the packages have been installed. So, remove it from here and L176 and just append it after the list of install statements has been generated.

lib/docker.js
```diff
@@ -135,24 +140,50 @@ function createDockerFile() {
...
+        extraPkgs.push('apt-transport-https');
     }
 
     contents = 'FROM ' + baseImg + '\n' +
-        'RUN apt-get update --fix-missing && apt-get install -y ' + extraPkgs.join(' ') +
+        'RUN apt-get update && apt-get install -y ' + extraPkgs.join(' ') +
         ' && rm -rf /var/lib/apt/lists/*\n';
 
+    customSourcePkgs.forEach(function(customSourcePkgSpec) {
+        contents += 'RUN echo "deb ' + customSourcePkgSpec.repo_url +
```
d00rman: Much better: compile the sources.list file first and then do apt-get update just once.

d00rman:
Very good work, @Pchelolo! Some minor comments in-lined and then we're GTG.

lib/docker.js
Pchelolo: @d00rman Unfortunately we can't do that, or I don't know how to achieve this. In `apt.wikimedia.org` the node version is 4.xxx and it depends on some npm version and gyp version which are not in jessie repo and not in apt.wikimedia repo, so if we try to install our normal deps after adding apt.wikimedia to the sources list - it fails. Would be awesome if you had an idea how to solve this.

d00rman: No no, I was thinking just of making an optimisation here: if there are multiple sources, add them all first, then update and install packages from these spurces. The packages from the main pool should always be installed before that precisely to avoid the problems you are describing.

Pchelolo: Aha, OK. I'll update the PR.

Pchelolo: @d00rman done

d00rman:
JSHint and jscs complain about some missing semicolons and indentation

lib/docker.js
```diff
@@ -135,24 +140,55 @@ function createDockerFile() {
...
     Object.keys(pkg.deploy.dependencies).forEach(function(sys) {
-        if (sys !== '_all' && (sys === baseImg || (new RegExp(sys)).test(baseImg))) {
-            Array.prototype.push.apply(extraPkgs, pkg.deploy.dependencies[sys]);
+        if (sys === '_all' || (sys === baseImg || (new RegExp(sys)).test(baseImg))) {
+            pkg.deploy.dependencies[sys].forEach(function(pkg) {
+                if (typeof pkg === 'string') {
+                    extraPkgs.push(pkg);
+                } else if (typeof pkg === 'object') {
+                    if (!pkg.repo_url || !pkg.pool || !pkg.packages || !pkg.release) {
+                        console.error('ERROR: Incorrect dependency spec: ' + JSON.stringify(pkg));
```
d00rman: The script should also probably `process.exit()` here.

lib/docker.js
```diff
@@ -556,14 +592,14 @@ function main(options, configuration) {
     return ensureDockerVersion()
     .then(getUid)
     .then(createDockerFile)
-    .then(buildImg)
+   /* .then(buildImg)
```
d00rman: Why is this commented out?

Pchelolo: Oh, that was a leftover from testing, sorry

d00rman:
Published as [v2.0.9](https://github.com/wikimedia/service-runner/tree/v2.0.9)

PR #119, opened 2016-08-16
Use default cluster scheduling policy & release v2.0.10

gwicke:


Pchelolo:
@gwicke So, you've finally decided going for it? Did you do some perf testing? Is the difference noticeable?

gwicke:
@Pchelolo: It's definitely noticeable in local testing with wrk, where I would otherwise occasionally see one or two workers take all the load (persistent connections). We are seeing similar load imbalances in prod, which this should help improve on.

Pchelolo:
@gwicke ok, let's give it a try then. merging

gwicke:
Effect on RB heap usage in prod: 

image

It's clear that load is now more evenly distributed across workers. The effect on client observed latency seems to be small for RB. It might be bigger for CPU-intensive services like Parsoid.

PR #120, opened 2016-08-18
Don't set two exit listeners simultationsly

Pchelolo:
When the worker starts up we set a specific exit listener, but also we set a cluster-wide exit listener. In some testing I managed to get worker in the state when it died during startup in the particular moment and both listeners fired spawning 2 workers instead of a single dead one. So instead of setting 2 listeners simultaneously replace the startup one with the normal one when worker reports it's startup finished properly.

Also there was a bug when worker didn't ever send any heartbeats master had an exception.

cc @wikimedia/services

### DIFF
--- a/lib/master.js
+++ b/lib/master.js
@@ -76,24 +76,6 @@ Master.prototype._start = function() {
     this._logger.log('info/service-runner', 'master(' + process.pid + ') initializing '
         + this.config.num_workers + ' workers');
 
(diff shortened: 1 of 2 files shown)


d00rman:
@pchelolo mind tagging and publishing? (not in a position to do it myself at the moment)

Pchelolo:
Done

PR #121, opened 2016-08-18
Ignore DNS resolution issues on logstash stream

Pchelolo:
Recent failure of change-prop was caused by this - we don't catch DNS resolution errors on self stream, so they end up in the unhandledException handler and kill the worker. Logging should not kill worker, so just ignore and hope other streams will deliver our message.

cc @wikimedia/services

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -51,15 +51,15 @@ function Logger(confOrLogger, args) {
 
 var streamConverter = {
     gelf: function(stream, conf) {
-        // Convert the 'gelf' logger type to a real logger
-        var gs = gelfStream.forBunyan(stream.host, stream.port, stream.options);
-        gs.on('error', function(err) {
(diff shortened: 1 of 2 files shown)


lib/logger.js
```diff
@@ -51,12 +51,15 @@ function Logger(confOrLogger, args) {
...
+            // Ignore, can't do anything, let's hope other streams will succeed
+        });
         // Convert the 'gelf' logger type to a real logger
         return {
             type: 'raw',
-            stream: gelfStream.forBunyan(stream.host,
-                stream.port, stream.options),
-            level: stream.level || conf.level
+            stream: impl,
+            level: impl.level || conf.level
```
d00rman: `stream.level || conf.level` ?

gwicke:
:+1:

d00rman:
Published as [v2.0.12](https://github.com/wikimedia/service-runner/tree/v2.0.12)

PR #122, opened 2016-08-19
Don't restart starting worker if shutting down

Pchelolo:
In case master is shutting down and the worker died exactly at that moment we shouldn't restart it.

Also got rid of `currentlyStartingWorker` property as we don't need it any more after we've stopped setting the global `cluster.on('exit')` callback

cc @wikimedia/services


d00rman:
Awesome find, @Pchelolo ! Thnx!

d00rman:
Published as [v2.0.13](https://github.com/wikimedia/service-runner/tree/v2.0.13)

PR #123, opened 2016-08-23
Allow env vars and per-service app_base_path config; release v2.1.0

d00rman:
When multiple services are declared in the config, they might not be able to have the same base path, so allow them to specify their own. The per-service base path does not have to be an absolute one, it can be relative to the global application base path.

Also, allow users to specify environment variables directly in `config.yaml` by using the construct
```yaml
field: '{env(ENV_VAR_NAME, defaultValue)}'
```

The replacement take place in the master process and then passed on to the workers. If an environment variable's value is not defined, the user can provide a default value.

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -183,6 +183,23 @@ BaseService.prototype._sanitizeConfig = function(conf, options) {
 };
 
 
+BaseService.prototype._replaceEnvVars = function(config) {
+    var envRegex = /\{\s*env\(([^,\s\)]+),?\s*([^\)]+)?\)\s*}/g;
+    if (Buffer.isBuffer(config)) {
+        config = config.toString();
+    }
+    return config.replace(envRegex, function(match, envName, defValue) {
+        if (process.env[envName] !== undefined) {
+            return process.env[envName];
+        }
+        if (defValue !== undefined) {
+            return defValue;
+        }
+        return '';
+    });
+};
+
+
 /**
  * Loads the config from file, serialized input or Object
  *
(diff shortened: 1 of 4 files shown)


d00rman:
cc @wikimedia/services

d00rman:
Hold on, I want to add another feature before releasing v2.1.0

lib/base_service.js
```diff
@@ -183,6 +183,22 @@ BaseService.prototype._sanitizeConfig = function(conf, options) {
 };
 
 
+BaseService.prototype._replaceEnvVars = function(config) {
```
Pchelolo: Would be nice to document this feature.

lib/base_service.js
```diff
@@ -183,6 +183,22 @@ BaseService.prototype._sanitizeConfig = function(conf, options) {
...
+    if(Buffer.isBuffer(config)) {
+        config = config.toString();
+    }
+    return config.replace(/\{\s*env\(([^,\s\)]+),?\s*([^\)]+)?\)\s*}/g, function(match, envName, defValue) {
```
Pchelolo: Which syntax are you using here? In bash the syntax is `{name:-default}` as far as I know. It that comma syntax coming from some standard?

d00rman: See the PR desc, it's `{env(VAR_NAME,defValue)}`

lib/base_service.js
d00rman: {{done}} in README.md

d00rman:
Ok, @wikimedia/services this is ready for review.

Pchelolo:
@d00rman travis...

d00rman:
@Pchelolo done, sorry about that

Pchelolo:
LGTM

Pchelolo:
Tagged and published as `v2.1.0`

PR #125, opened 2016-09-12
Added function to rateLimiter to check rate limit

Pchelolo:
I need this in Change-Propagation to improve how our rate-limiting works.

cc @wikimedia/services


d00rman:
Published as [v2,1.2](https://github.com/wikimedia/service-runner/tree/v2.1.2)

PR #126, opened 2016-09-13
Docker: Allow the creation of non-unique uid/gid inside the container

d00rman:


d00rman:
cc @wikimedia/services

Pchelolo:
It's good, so merging, just a side note - this shouldn't be required to support docker scripts on OS X. But it wouldn't hurt.

Pchelolo:
Tagged and published.

PR #128, opened 2016-09-13
Adopt yargs@5.0.0

Pchelolo:
Follow-up, let's go back to args@5.0.0. I thought they've had a bug, but that's just not that well documented breaking change. No need to release a new version, nothing is changed really

cc @wikimedia/services

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -77,52 +77,59 @@ BaseService.prototype._getOptions = function(opts) {
     // check process arguments
     var args = require('yargs')
     .usage('Usage: $0 [command] [options]')
-    .command('docker-start', 'starts the service in a Docker container')
-    .command('docker-test', 'starts the test process in a Docker container')
-    .command('build', 'builds the service\'s package and deploy repo')
     .options({
         n: {
             alias: 'num-workers',
             default: -1,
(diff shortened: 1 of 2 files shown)


lib/base_service.js
```diff
@@ -85,43 +85,50 @@ BaseService.prototype._getOptions = function(opts) {
...
+            type: 'boolean',
+            global: true
         },
         d: {
             alias: 'deploy-repo',
             default: false,
             describe: 'build only the deploy repo',
-            type: 'boolean'
+            type: 'boolean',
+            global: true
```
d00rman: Euh, this should pertain only to the build command, shouldn't it?

lib/base_service.js
```diff
@@ -85,43 +85,50 @@ BaseService.prototype._getOptions = function(opts) {
...
         },
         r: {
             alias: 'review',
             default: false,
             describe: 'send the patch to Gerrit after building the repo',
-            type: 'boolean'
+            type: 'boolean',
+            global: true
```
d00rman: Idem.

d00rman:
Theoretically `--force` should be a global option, but given the current usage, it doesn't matter much. Nice work @Pchelolo !

PR #129, opened 2016-09-14
Support OSX-style find command

Pchelolo:
OSX has different 'find' syntax. 

cc @wikimedia/services

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -451,17 +451,18 @@ function updateDeploy() {
             return startContainer(['-v', opts.dir + ':/opt/service', '-w', '/opt/service'], true);
         }).then(function() {
             // remove unnecessary files
+            var findAttr;
             if (os.type() === 'Darwin') {
-                return promisedSpawn([
+                findAttr = [
                     'find',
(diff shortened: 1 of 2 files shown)


lib/docker.js
```diff
@@ -451,15 +451,26 @@ function updateDeploy() {
...
-            return promisedSpawn([
-                'find',
-                'node_modules/',
-                '-regextype',
-                'posix-egrep',
-                '-iregex',
-                '(.*\\.git.*|.*\\.md|.*readme|.*licence)',
-                '-exec', 'rm', '-rf', '{}', ';'
-            ], { capture: true, ignoreErr: true });
+            if (os.type() === 'Darwin') {
```
d00rman: A minor style nit: how about setting an array based on the OS and then doing

```js
return promisedSpawn(findArr, { capture: true, ignoreErr: true });
```

IMHO, it'd make the code more legible.

d00rman:
Published as [v2.1.5](https://github.com/wikimedia/service-runner/tree/v2.1.5)

PR #130, opened 2016-09-15
Docker: Allow users to specify src and deploy repo branches to use; release v2.1.6

d00rman:


d00rman:
cc @wikimedia/services

d00rman:
Self-merging as it is needed for the AQS deploy

d00rman:
Published as [v2.1.6](https://github.com/wikimedia/service-runner/tree/v2.1.6). [The related ocumentation](https://www.mediawiki.org/wiki/ServiceTemplateNode/Deployment#Local_git) has been updated as well.

PR #131, opened 2016-09-15
POC implementation of sticky cluster

Pchelolo:
WIP, proof of concept.

Problem: socket.remoteAddr is actually not we want, it will probably point to some load balancer, we want 'X-Client-IP' header, but we can't get it here since as I understand we can't read the data in master if we wanna be able to pass the socket handle around.


Pchelolo:
We don't need this after all.

PR #133, opened 2016-09-26
Support providing multiple servers in gelf config

Pchelolo:
One of the blockers in moving RESTBase to the generic puppet service module is the fact that RESTBase and other services use different logstash hosts. Support providing a list of logstash hosts in the config and randomly selecting one on startup, that would end up in better load-balancing for logstash.

cc @wikimedia/services

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -51,7 +51,17 @@ function Logger(confOrLogger, args) {
 
 var streamConverter = {
     gelf: function(stream, conf) {
-        var impl = gelfStream.forBunyan(stream.host, stream.port, stream.options);
+        var host = stream.host;
+        var port = stream.port;
+        if (stream.uris) {
+            var hosts = stream.uris.split(',');
+            var selectedHost = hosts[Math.floor(Math.random() * hosts.length)];
+            host = selectedHost.substr(0, selectedHost.indexOf(':'));
+            port = parseInt(selectedHost.substr(selectedHost.indexOf(':') + 1));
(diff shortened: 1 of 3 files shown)


lib/logger.js
```diff
@@ -51,7 +51,17 @@ function Logger(confOrLogger, args) {
...
 var streamConverter = {
     gelf: function(stream, conf) {
-        var impl = gelfStream.forBunyan(stream.host, stream.port, stream.options);
+        var host = stream.host;
+        var port = stream.port;
+        if (stream.uris) {
+            var hosts = stream.uris.split(',');
+            var selectedHost = hosts[Math.floor(Math.random() * hosts.length)];
+            host = selectedHost.substr(0, selectedHost.indexOf(':'));
+            port = parseInt(selectedHost.substr(selectedHost.indexOf(':') + 1));
```
d00rman: Should add a fallback - either `stream.port` or a sane default.

lib/logger.js
```diff
@@ -51,7 +51,17 @@ function Logger(confOrLogger, args) {
...
+        }
+
+        var impl = gelfStream.forBunyan(host, port, stream.options);
+        console.log(host, port);
```
d00rman: Ups?

Pchelolo:
@d00rman Updated

d00rman:
Published as [v2.1.7](https://github.com/wikimedia/service-runner/tree/v2.1.7)

PR #134, opened 2016-10-27
Added --reshrinkwrap to support shrinkwrapping

nyurik:


Pchelolo:
LGTM. Let's wait for travis though

Pchelolo:
@nyurik The JSCS check is failing. Please fix and I'll merge and publish

d00rman:
I may be missing something, but what is the point of regenerating that file? If this were to be merged, then functionally there is no difference between using the new switch and not using shrinkwrap altogether.

Shouldn't we strive towards actually making use of it during the install process instead? One could have it in the source repo and then link it in the deploy repo, much like we do now for `package.json` and have npm obey it.

nyurik:
@d00rman I find this switch highly needed in the following scenario:
* My code relies on https://github.com/mapnik/node-mapnik lib, but I must restrict it to a specific version due to the newer CLIB requirement which we don't yet support in production
* My deploy repo contains a `npm-shrinkwrap.json` file
* Every time I need to build a new deploy repo, I do these steps:
** `./server.js build --deploy-repo --force --rebuild-shrinkwrap`
** edit `npm-shrinkwrap.json` to use the needed Mapnik lib
** reset deploy repo back to the current master
** commit deploy repo with the updated `npm-shrinkwrap.json`
** Rebuild deploy repo without the rebuild switch to use specific dependencies versions: `./server.js build --deploy-repo --force`

d00rman:
@nyurik I think what would the correct way to do it is to have the shrinkwrap json ready in the source repo and link it in the deploy repo. That way, when you update the src pointer there, the correct shrinkwrap json will be read by `npm install`. It's a good point about `npm dedupe` - it should not be used if `npm-shrinkwrap.json` is present. I will fix that. This combined with symlinking that file to the source repo should get you what you want and you will be able to execute the build script only once.

Also, doesn't your code require mapnik directly? If so, can't you fix its version in your `package.json`?

nyurik:
@d00rman the source repo does not need it because limiting mapnik is a production requirement, not a development requirement. Moreover, I think I will eventually have to have a separate version of package.json - because production uses a specific set of dependencies (plugins), which is very different from what others may want to use in their setup. (I need to eventually switch to a proper dependency injection framework, that should make it simpler).

I prefer to do npm shrinkwrap from insider docker because it produces a much cleaner output, without any extra dev dependencies. Also, I think dedup is highly needed - I would end up with tons of duplicates otherwise. Also, I do not want to manually update shrinkwrap file every time some dependency of dependency changes - I prefer to run "npm install" as if there is no shrinkwrap, fix just the minor issue with Mapnik, and shrinkwrap it in place.

My code doesn't actually require Mapnik, but many plugins do. Plus it is very important that there are no two different versions by accident (which has happened before) - because its a native lib, it crashes with two vers.

d00rman:
Unfortunately, we'll have to get rid of `npm dedupe` when `npm-shrinkwrap.json` is present, as that blocks others that rely on this file to use service-runner's build process. I created [T149979](https://phabricator.wikimedia.org/T149979) for this.

I think it's bad practice not to have the same versions in development and production, as this creates incompatibility problems, as you've noticed yourself. I would urge you to reconsider this practice. Others that use the repo would surely prefer to use a stable version, and those that really want bleeding-edge versions (which is always a minority) can install them themselves.

> I prefer to do npm shrinkwrap from insider docker because it produces a much cleaner output, without any extra dev dependencies.

You can easily achieve this by running `npm install --production && npm shrinkwrap && npm install`. Then, the dev dependencies will not end up inside the file.

d00rman:
Published as [v2.1.8](https://github.com/wikimedia/service-runner/tree/v2.1.8)

PR #137, opened 2016-11-10
Provide the worker id to the service

Pchelolo:
In the trending service we would need to construct a stable consumer group id unique for each worker. We could use the `os.hostname()` to differentiate the workers between the hosts, but for workers on the same host we need a stable unique id. A sequential number of the worker in startup sequence could work as such an id, so forward it to the service together with the service configuration.

This problem could've been solved my having just one worker per host, but in case the worker dies, spawning the new one will take quite some time, and in the meantime all requests to a particular host would fail. So it's better to be able to have 2 workers per node.

cc @wikimedia/services

### DIFF
--- a/lib/master.js
+++ b/lib/master.js
@@ -194,7 +194,7 @@ Master.prototype._onStatusReceived = function(worker, status) {
 
 // Fork off one worker at a time, once the previous worker has finished
 // startup.
-Master.prototype._startWorkers = function(remainingWorkers, res) {
+Master.prototype._startWorkers = function(remainingWorkers, res, worker_id) {
     res = res || [];
     var self = this;
     if (remainingWorkers) {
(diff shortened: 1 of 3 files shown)


lib/master.js
```diff
@@ -202,11 +203,13 @@ Master.prototype._startWorkers = function(remainingWorkers, res) {
         self._saveBeat(worker);
         return new P(function(resolve) {
             fixCloseDisconnectListeners(worker);
+            var config = Object.assign({}, self.config, {
+                worker_id: self.config.num_workers - remainingWorkers + 1
```
gwicke: Won't restarted workers all end up sharing the same worker id eventually?

Pchelolo: Oh you're right.. That was stupid, will update

Pchelolo: @gwicke Fixed, now IDs are stable across worker's death.

gwicke:
LGTM. I'm still not a fan of adding all this complexity to work around the temporary lack of time-based indexing in eventstreams, but that's a different discussion.

PR #138, opened 2016-11-11
Allow installing custom deb packages

Pchelolo:
For kartoterian support on node 6 and jessie for now we need to install libmapnik packages Faidon has built. Add support for installing custom deb packages from some URIs.

We might wanna delete this feature eventually, but for now there's no real way around

cc @wikimedia/services @nyurik

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -123,6 +123,8 @@ function createDockerFile() {
     // 'packages' is an array of package names
     // 'repo_url' and 'pool' combinations are unique
     var customSourcePkgs = [];
+    // An array of URIs of custom .deb packages to be installed
+    var debPkgs = [];
 
     // set some defaults
     if (!pkg.deploy) {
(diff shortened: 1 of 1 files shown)


nyurik:
@pchelolo, thanks! How would one use it? Is there a fix for it?

d00rman:
LGTM, but needs a couple of fixes jscs is complaining about.

@nyurik this will be available in the next version of service-runner. You can add the deb's URI directly in package.json like so:

```json
{
  "deploy": {
    "dependencies": {
      "debian": [
        "<complete_uri_to_deb>"
      ]
    }
  }
}
```

lib/docker.js
```diff
@@ -153,11 +155,16 @@ function createDockerFile() {
                 if (typeof pkg === 'string') {
                     extraPkgs.push(pkg);
                 } else if (typeof pkg === 'object') {
-                    if (!pkg.repo_url || !pkg.pool || !pkg.packages || !pkg.release) {
-                        console.error('ERROR: Incorrect dependency spec: ' + JSON.stringify(pkg));
-                        process.exit(1);
+                    if (pkg.uri) {
+                        debPkgs.push(pkg.uri);
```
gwicke: [Minor] naming suggestion: `debPackageURLs`

PR #139, opened 2016-11-11
Bump version to 2.1.9

Pchelolo:


Pchelolo:
Self-merging, just a version bump

PR #140, opened 2016-11-11
Allow additional install options in the build script

Pchelolo:
For kartotherian we need to provide --build-from-source=mapnik --fallback-to-build=false options to the `npm install` command, allow that.

cc @wikimedia/services @nyurik

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -247,7 +247,11 @@ function createDockerFile() {
             beforeInstall = 'rm -f npm-shrinkwrap.json && ';
             afterInstall = '&& npm shrinkwrap ';
         }
-        contents += 'CMD ' + beforeInstall + npmCommand + ' install --production ' + afterInstall +
+        var installOpts = ' --production ';
+        if (pkg.deploy.install_opts) {
+            installOpts += pkg.deploy.install_opts.join(' ') + ' ';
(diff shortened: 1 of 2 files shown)


gwicke:
Hmmm... the number of install options keeps getting growing. Perhaps that build.sh idea wasn't so bad after all?

Pchelolo:
@gwicke Rebased. Let's support this by docker script, and if anything else comes up - then let's move to build.sh

d00rman:
LGTM, but could you also add the docs for it? As @gwicke points out, we have a lot of stuff now that can go into the `deploy` section of `package.json`, so let's keep our sanity, at least partially :)

Pchelolo:
@d00rman Wrote some docs on https://www.mediawiki.org/wiki/ServiceTemplateNode/Deployment

d00rman:
LGTM, but both this PR and #142 bump the pkg version. They could both go in the same version. Could you remove the bump from one of the two and we can then merge both and publish.

Pchelolo:
@d00rman Rebased. We're still gonna need this.

d00rman:
Published as (v2.1.15)[https://github.com/wikimedia/service-runner/tree/v2.1.15]

PR #141, opened 2016-11-15
Don't dedupe if shrinkwrap is present

Pchelolo:
Bug: https://phabricator.wikimedia.org/T149979
cc @wikimedia/services


gwicke:
LGTM.

PR #142, opened 2016-11-21
Added support for specifying the repo key

Pchelolo:
@paravoid correctly pointed out that using `--force-yes` when installing packages from WMF apt repo is a terrible practice (https://gerrit.wikimedia.org/r/#/c/322368/), so add support for providing key URL so that we could fetch the key to the repo and properly add it to apt-get

cc @wikimedia/services

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -107,6 +107,15 @@ function promisedSpawn(args, options) {
 
 }
 
+function generateCustomSourcesKeyFetch(customSourcePkgs) {
+    customSourcePkgs = customSourcePkgs.filter(function(pkgSpec) { return pkgSpec.key_url; });
+    if (customSourcePkgs.length) {
+        return 'RUN ' + customSourcePkgs.map(function(pkgSpec) {
+                return ' wget -O key "' + pkgSpec.key_url + '" && apt-key add key && rm key ';
+            }).join('&&') + '\n';
+    }
+    return '';
+}
 
 /**
  * Generates the Dockerfile used to build the image and start the container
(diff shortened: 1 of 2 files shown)


Pchelolo:
Will update the documentation once it's merged/

lib/docker.js
```diff
@@ -107,6 +107,15 @@ function promisedSpawn(args, options) {
 
 }
 
+function generateCustomSourcesKeyFetch(customSourcePkgs) {
+    customSourcePkgs = customSourcePkgs.filter(function(pkgSpec) { return pkgSpec.key_url; });
+    if (customSourcePkgs.length) {
+        return 'RUN ' + customSourcePkgs.map(function(pkgSpec) {
+                return ' wget -O key "' + pkgSpec.key_url + '" && apt-key add key && rm key ';
```
pnorman: can the output of wget be piped to apt-key to avoid the temp file?

d00rman: It could, but since we are doing this inside a container and already need to use temporary files and caches for fetching and installing DEBs anyway, I don't think this matters in practice.

paravoid:
You should reference a key fingerprint, not a key URL. Even if the Wikitech page was authoritative (it's not, anyone can change that), referencing a URL is not more secure than just passing --force-yes; wikitech.wm.org is not more secure than apt.wm.org (https helps but it's still not a great idea to delegate cryptographic checks to the contents of a URL).

Also, I know it probably doesn't matter here but I can't help but notice the temp file vulnerability due to the predictably-named temporary file :)

Pchelolo:
@paravoid Hm, do we have a key server to get the key from via hkp? [Wikitech page](https://wikitech.wikimedia.org/wiki/APT_repository) says nothing about it.. Or you mean I should still download the key by wget but additionally verify the fingerprint afterwards?

d00rman:
@paravoid ping re @Pchelolo's question above ^

PR #144, opened 2016-12-09
Added ability for lazy log payload creation

Pchelolo:
After playing with node 6 new feature `--inspect` and profiling change-prop a bit I've realised that about 10% of the CPU is spent in constructing logging statements, because we need to stringify events to avoid logstash issues. But most of the logs never actually make it, so all this effort is lost. 

The idea is to accept a callback function to lazily construct the log payload after we've decided that it's needed. Other option was to create an `isLoggable` method and add an `if` statement on the call, but figuring out if the level is loggable is a pretty expensive process so we don't want to repeat it twice.

cc @wikimedia/services

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -296,27 +296,42 @@ Logger.prototype._log = function(info, level, levelPath, logger) {
     }
 };
 
-Logger.prototype.log = function(level, info) {
+/**
+ * Logs and info object with a specified level
+ * @param {string} level Log level and components, for example 'trace/request'
+ * @param {Object|Function} info log statement object, or a callback to lazily construct the log statement
+ *                               after we've decided that this particular level will be matched.
+ * @param {boolean} [force] optional argument to force sampled component logging even
+ *                          if the probability wasn't met. Used together with `isLoggable`
+ */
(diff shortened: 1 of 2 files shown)


d00rman:
> about 10% of the CPU is spent in constructing logging statements, because we need to stringify events

But stringification happens only once we sure that the log entry need to be sent. I don't understand the difference here - regardless of whether one passes a function returning an object or the object itself, they get evaluated only once we know that we should send the log. Or am I missing something?

Pchelolo:
@d00rman Compare this:
```javascript
log('trace/request', {
  value: JSON.stringify(event)
});
```
to this:
```javascript
log('trace/request', () => ({
  value: JSON.stringify(event)
}));
```
In the first example stringification happens BEFORE we pass it to the log function and then if the level is not met it's thrown away. In the second example stringification happens only AFTER we decided we need to log and use the value.

d00rman:
Oh right, that stringification. Makes sense now!

d00rman:
Published as [v2.1.13](https://github.com/wikimedia/service-runner/tree/v2.1.13)

PR #146, opened 2017-02-07
Docker: Allow only the Dockerfile to be generated

d00rman:
As a first step of the containerasition effort, we need to make the
Dockerfile specification based on package.json available to other
parts of the system, most notably the build and test pipeline(s). To
that end, service-runner needs to be able to only generate the spec
without actually building nor starting the container.

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -109,33 +109,51 @@ BaseService.prototype._getOptions = function(opts) {
     })
     .command('docker-start', 'starts the service in a Docker container')
     .command('docker-test', 'starts the test process in a Docker container')
-    .command('build', 'builds the service\'s package and deploy repo', function(yargs) {
-        return yargs.option({
-            f: {
-                alias: 'force',
-                default: false,
-                describe: 'force the operation to execute',
-                type: 'boolean'
-            },
-            d: {
-                alias: 'deploy-repo',
-                default: false,
-                describe: 'build only the deploy repo',
-                type: 'boolean'
-            },
-            s: {
-                alias: 'reshrinkwrap',
-                default: false,
-                describe: 'rebuild shrinkwrap.json by removing and regenerating after npm install',
-                type: 'boolean'
-            },
-            r: {
-                alias: 'review',
-                default: false,
-                describe: 'send the patch to Gerrit after building the repo',
-                type: 'boolean'
-            }
-        });
+    .command('build', 'builds the service\'s package and deploy repo', {
+        f: {
+            alias: 'force',
+            default: false,
+            describe: 'force the operation to execute',
+            type: 'boolean'
+        },
+        d: {
+            alias: 'deploy-repo',
+            default: false,
+            describe: 'build only the deploy repo',
+            type: 'boolean'
+        },
+        s: {
+            alias: 'reshrinkwrap',
+            default: false,
+            describe: 'rebuild shrinkwrap.json by removing and regenerating after npm install',
+            type: 'boolean'
+        },
+        r: {
+            alias: 'review',
+            default: false,
+            describe: 'send the patch to Gerrit after building the repo',
+            type: 'boolean'
+        }
+    })
+    .command('generate', 'generates the Dockerfile specification for the service', {
+        r: {
+            alias: 'running',
+            default: false,
+            describe: 'generate the Dockerfile to start the service',
+            type: 'boolean'
+        },
+        t: {
+            alias: 'testing',
+            default: false,
+            describe: 'generate the Dockerfile to test the service',
+            type: 'boolean'
+        },
+        b: {
+            alias: 'building',
+            default: false,
+            describe: 'generate the Dockerfile to build the deployment repository',
+            type: 'boolean'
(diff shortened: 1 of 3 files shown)


d00rman:
cc @wikimedia/services @wikimedia/release-engineering

lib/base_service.js
```diff
@@ -107,35 +107,50 @@ BaseService.prototype._getOptions = function(opts) {
...
+            describe: 'rebuild shrinkwrap.json by removing and regenerating after npm install',
+            type: 'boolean'
+        },
+        r: {
+            alias: 'review',
+            default: false,
+            describe: 'send the patch to Gerrit after building the repo',
+            type: 'boolean'
+        },
+        nobuild: {
```
Pchelolo: I think that could be a new high-level command `generate`. The command `service-runner build --nobuild` looks incredibly confusing..

lib/base_service.js
```diff
@@ -107,35 +107,50 @@ BaseService.prototype._getOptions = function(opts) {
...
-        });
+    .command('docker-start', 'starts the service in a Docker container', {
+        nobuild: {
+            default: false,
+            describe: 'do not build nor execute the container, only generate the Dockerfile spec',
+            type: 'boolean'
+        }
+    })
+    .command('docker-test', 'starts the test process in a Docker container', {
+        nobuild: {
```
Pchelolo: What would the command `service-runner docker-test --nobuild` do? How would it run the tests if it's not building the container? I'm confused.

hashar: The use case we have is to only get the DockerFile generated, though I guess one could also be interested in solely building the image and not running anything.   What about having top level commands such as:

 service-runner docker-file   # would generate the DockerFile
 service-runner docker-build  # generate DockerFile + docker build
 service-runner docker-start
 service-runner docker-test

Namely factors out the DockerFile generation and image building.

d00rman: IMHO, `docker-build` just complicates things because it creates confusion with the existing `build` command, and building the container after having the Dockerfile is a one-liner, so it can be easily scripted away.

d00rman:
@Pchelolo updated

lib/base_service.js
hashar: Yeah indeed that is confusing. The latest patch with 'generate testing'  looks human friendly enough :]

lib/base_service.js
```diff
@@ -145,13 +163,16 @@ BaseService.prototype._getOptions = function(opts) {
...
     args.dockerStart = args._.indexOf('docker-start') !== -1;
     args.dockerTest = args._.indexOf('docker-test') !== -1;
-    args.deployRepo = args.deployRepo || args.build;
+    args.generate = args._.indexOf('generate') !== -1;
+    args.deployRepo = args.deployRepo || args.build || args.building;
 
-    if (args.build && args.dockerStart || args.build && args.dockerTest
-    || args.dockerStart && args.dockerTest) {
+    /*jshint bitwise: false*/
+    if (!(args.build ^ args.dockerStart ^ args.dockerTest ^ args.generate) &&
```
gwicke: Perhaps more obvious: `[args.build, args.dockerStart, args.dockerTest, args.generate].filter(x => x).length > 1`

d00rman: @gwicke Eh, we still haven't deprecated node 0.1x for service-runner so that's not going to fly, but I agree it might be slightly obvious (even though in both cases one needs to think what's going on). Luckily the next line explains it pretty clearly :)

gwicke: Well, you can just use a normal function then.

d00rman: do you feel strongly about this? It seems like a nit-pick to me.

d00rman: It doesn't seem as natural with the old syntax to me, it's clobbering it imho

gwicke: Sure, it is a nit-pick & doesn't change functionality. Readability is important, though, and imho the filter version is both shorter and more obvious. It also avoids setting off jshint.

gwicke: Tested with 0.10: `[true, false, false, false].filter(Boolean).length`

d00rman:
good to go now.

lib/base_service.js
```diff
@@ -145,10 +163,11 @@ BaseService.prototype._getOptions = function(opts) {
...
     args.dockerStart = args._.indexOf('docker-start') !== -1;
     args.dockerTest = args._.indexOf('docker-test') !== -1;
-    args.deployRepo = args.deployRepo || args.build;
+    args.generate = args._.indexOf('generate') !== -1;
+    args.deployRepo = args.deployRepo || args.build || args.building;
 
-    if (args.build && args.dockerStart || args.build && args.dockerTest
-    || args.dockerStart && args.dockerTest) {
+    if ([args.build, args.dockerStart, args.dockerTest, args.generate]
+            .filter(function(x) { return !!x; }).length()) {
```
gwicke: `.length` is a property, not a function.

gwicke:
:+1:

PR #147, opened 2017-02-08
Docker: Fetch all updates

d00rman:
Developers can have multiple remotes defined for the deploy repository.
Additionally, newer versions of git-review automatically create a
'gerrit' remote that is equal to 'origin', so fetch all of the updates
to avoid confusing gerrit when submitting the patch.


d00rman:
Trivial, so self-merged.

PR #148, opened 2017-02-08
Release v2.1.14 and update dependencies

d00rman:


d00rman:
Published as [v2.1.14](https://github.com/wikimedia/service-runner/tree/v2.1.14)

PR #149, opened 2017-02-13
Docker: No need to check for Docker if it will not be used

d00rman:
As of recently, service-runner is able to only generate the Dockerfile
specification without actually using it. In this case, there is no need
to fail if Docker is not installed on the system generating the
specification, since it will never be invoked.


d00rman:
cc @wikimedia/services @wikimedia/release-engineering

gwicke:
Trivial enough to ignore the tardy Travis guy this time.

PR #150, opened 2017-02-16
Cleanup timers so process shuts down cleanly

arlolra:
See https://github.com/gwicke/limitation/pull/1

Part of T158248

/cc @wikimedia/services

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -31,6 +31,15 @@ function BaseService(options) {
     this.serviceReturns = null;
 }
 
+BaseService.prototype.stop = function() {
+    if (this.interval) {
+        clearInterval(this.interval);
+    }
+    if (this._ratelimiter) {
+        this._ratelimiter.stop();
+    }
+};
+
 BaseService.prototype._setAppBasePath = function(config) {
     if (process.env.APP_BASE_PATH) {
         this._basePath = process.env.APP_BASE_PATH;
--- a/lib/master.js
+++ b/lib/master.js
@@ -51,11 +51,9 @@ Master.prototype.stop = function() {
         // No workers needed, run worker code directly.
         return Worker.prototype.stop.call(self);
     }
+    BaseService.prototype.stop.call(self);
     self._shuttingDown = true;
     self._logger.log('info/service-runner/master', 'master shutting down, killing workers');
-    if (self.interval) {
-        clearInterval(self.interval);
-    }
     return P.map(Object.keys(cluster.workers), self._stopWorker.bind(self));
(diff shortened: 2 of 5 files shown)


Pchelolo:
Heh, I've been doing pretty much the same thing right now as a part of looking into https://phabricator.wikimedia.org/T158265

arlolra:
@Pchelolo Ok, this was T158248.  If you've got something more fully baked though, use that and drop this.

lib/base_service.js
```diff
@@ -31,6 +31,15 @@ function BaseService(options) {
     this.serviceReturns = null;
 }
 
+BaseService.prototype.cleanup = function() {
```
Pchelolo: Could we call this method also `stop` and then override it in Master and Worker and call a superclass method with something like `BaseService.prototype.cleanup.call(this)`? It's gonna be cleaner in OOP sense and also when we switch to class syntax it should be nicer. Not a strong opinion though

Pchelolo:
Let's make a version bump.

package.json
```diff
@@ -35,7 +35,7 @@
     "gelf-stream": "^1.1.1",
     "hot-shots": "^4.3.1",
     "js-yaml": "^3.8.1",
-    "limitation": "^0.1.9",
+    "limitation": "git+https://github.com/arlolra/limitation#cleanup",
```
Pchelolo: I will review the `limitation PR`, we'll bump a version there and replace this with normal dependency before merging.

lib/base_service.js
arlolra: Sounds good.

package.json
arlolra: Yup, I did this so tests would pass ... and then there weren't any tests :P

Pchelolo: @arlolra Ye, we know, shame on us, there's no tests for `service-runner`, but as an excuse I would say it would be extremely hard to write unit tests for this..

lib/master.js
```diff
@@ -53,9 +53,7 @@ Master.prototype.stop = function() {
     }
     self._shuttingDown = true;
     self._logger.log('info/service-runner/master', 'master shutting down, killing workers');
-    if (self.interval) {
-        clearInterval(self.interval);
-    }
+    self.cleanup();
```
arlolra: This needs to move higher up, too.

arlolra:
Updated to reflect the changes to the dependency.

Pchelolo:
Published!

arlolra:
Thanks.

PR #152, opened 2017-02-17
Support DNS caching

Pchelolo:
Added support for DNS caching and disabled internal statsd DNS caching. 

This will have quite a brutal effect on performance if we just disable internal statsd client cache and not switch on the global cache, so I've bumped the package version to 2.2.0.

Bug: https://phabricator.wikimedia.org/T158338
cc @wikimedia/services

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,11 @@ ratelimit:
   #  - 123.456.789.233
   #  - 456.789.90.2
 
+# DNS caching settings
+dns_cache:
+  ttl: 5 # Time-to-live for cache entries, in seconds
+  size: 10 # Optional cache size, default: 100
+
 services:
   - name: parsoid
     # a relative path or the name of an npm package, if different from name
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ ratelimit:
 # DNS caching settings
 dns_cache:
   ttl: 5 # Time-to-live for cache entries, in seconds
-  size: 10 # Optional cache size, default: 100
+  size: 100 # Optional cache size, default: 100
 
 services:
   - name: parsoid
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -157,6 +157,11 @@ ratelimit:
   #  - 123.456.789.233
   #  - 456.789.90.2
 
+# DNS caching settings
+dns_cache:
+  ttl: 5 # Time-to-live for cache entries, in seconds
+  size: 10 # Optional cache size, default: 100
```
gwicke: Did you see a strong reason for setting such a small cache size?

Pchelolo: It's just an example, the default is 100, I don't think any of the services use > 100 domains..

gwicke: I see. Since examples are often copy / pasted around, lets perhaps keep it at 100, to avoid surprises with bad performance?

gwicke:
LGTM overall.

README.md
Pchelolo: Good point. Updated to 100

Pchelolo:
> If we ever want to improve DNS performance further, then I think we could consider re-implementing the cache class in dnscache with something that's not strictly LRU (timer driven cleanup, amortized across many lookups), and updates popular entries asynchronously without blocking ongoing requests.

That could be interesting, but I don't think we should do it prematurely. My local testing shows that performance impact is very low even with switching from infinite caching to 5-second TTL, it reality it might be higher though.

gwicke:
It shouldn't affect mean throughput much, but especially async updates and perhaps a grace TTL extension in case of temporary resolution failures might be noticeable for p99. We had outages caused by temporary DNS resolution wobbles before, and it should be possible to paper those over with a smarter cache.

PR #154, opened 2017-02-21
Correctly check whether npm-shrinkwrap exists

Pchelolo:
Ok, before if the file existed docker would exit with code 1 breaking the build.

cc @wikimedia/services

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -254,7 +254,7 @@ function createDockerFile() {
         contents += 'CMD ' + beforeInstall + npmCommand + ' install' + installOpts + afterInstall +
             ' && npm install heapdump ';
         if (!opts.reshrinkwrap) {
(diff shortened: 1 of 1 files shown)


lib/docker.js
```diff
@@ -254,7 +254,7 @@ function createDockerFile() {
         contents += 'CMD ' + beforeInstall + npmCommand + ' install' + installOpts + afterInstall +
             ' && npm install heapdump ';
         if (!opts.reshrinkwrap) {
-            contents += '&& ! [ -e npm-shrinkwrap.json ] && npm dedupe ';
+            contents += '&& if ! [ -e npm-shrinkwrap.json ] ; then npm dedupe ; fi';
```
gwicke: Interesting nesting of expressions and `&&`.

PR #156, opened 2017-02-22
Don't bring down the service if an error occurs in the logger

Pchelolo:
I've been testing what happens when the disk partition we log to fills up. Currently that exception is not 
being handled, and the whole service (including master process) just dies. 

We don't want to bring down the whole service because of the logger, but we as well can't really handle properly any fatal logging errors, so let's just ignore them hoping that if one stream failed, other streams would succeed.

However, disk fill-up is a different situation. In that case every attempt to log something would result in a memory leak, quickly exhausting the process memory. To fix it up, we can remove the file logging stream and hope that some other streams would we able to deliver our error message.

Bug: https://phabricator.wikimedia.org/T112648

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -22,8 +22,13 @@ function Logger(confOrLogger, args) {
         self._sampled_levels = conf.sampled_levels || {};
         delete conf.sampled_levels;
         self._logger = bunyan.createLogger(conf);
+        self._logger.on('error', function() {
+            // If some fatal error occurs in the logger,
+            // we cannot really do anything, let's just hope
+            // some of the streams succeed, but definitely
+            // don't bring down the service.
+        });
         self._levelMatcher = this._levelToMatcher(conf.level);
-
         // For each specially logged component we need to create
         // a child logger that accepts everything regardless of the level
         self._componentLoggers = {};
@@ -32,10 +37,16 @@ function Logger(confOrLogger, args) {
                 component: component,
                 level: bunyan.TRACE
             });
+            self._componentLoggers[component].on('error', function() {
+                // Don't bring down the service if can't log
+            });
         });
 
         self._traceLogger = self._logger.child({
             level: bunyan.TRACE
+        });
+        self._traceLogger.on('error', function() {
+            // Don't bring down the service if can't log
         });
         // Set up handlers for uncaught extensions
         self._setupRootHandlers();
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -22,12 +22,7 @@ function Logger(confOrLogger, args) {
         self._sampled_levels = conf.sampled_levels || {};
         delete conf.sampled_levels;
         self._logger = bunyan.createLogger(conf);
-        self._logger.on('error', function() {
-            // If some fatal error occurs in the logger,
-            // we cannot really do anything, let's just hope
-            // some of the streams succeed, but definitely
-            // don't bring down the service.
-        });
+        self._addErrorHandler();
         self._levelMatcher = this._levelToMatcher(conf.level);
         // For each specially logged component we need to create
         // a child logger that accepts everything regardless of the level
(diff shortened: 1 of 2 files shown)


lib/logger.js
```diff
@@ -121,6 +121,29 @@ var streamConverter = {
...
+    self._logger.on('error', function(err, failedStream) {
+        // If some fatal error occurred in one of the logger streams,
+        // we can't do much, so ignore.
+
+        if (err.code === 'ENOSPC') {
+            // However, if it's a `file` stream, and the error is from
+            // a filled-up disc, attempting to further write to this
+            // stream will leak memory and finally destroy the process.
+            // So remove destroy the stream and remove it.
+            self._logger.streams = self._logger.streams.filter(function(s) {
```
d00rman: Does the filtered-out stream get preserved on SIGHUPs?

Pchelolo: @d00rman Eeeeemmm... In master we recreate the logger on SIGHUP, so the failed stream comes back. Worker processes get recreated completely, so also, stream comes back.

d00rman: Hm, how about we note that somewhere and don't create `file`-type log resources since a SIGHUP might still happen in that case due to log rotation?

lib/logger.js
```diff
@@ -121,6 +121,29 @@ var streamConverter = {
...
+                return s !== failedStream;
+            });
+            failedStream.stream.destroy();
+            // Hope that we have other streams to report the problem
```
d00rman: What about checking for the length of the array and remove the loggers altogether if it's empty? Better yet, have a `null` logger in that case?

lib/logger.js
d00rman: Oh, but that might be tricky actually. What if we have multiple `file`-type loggers that log to different partitions and only one is saturated? hm... I guess the worst-case scenario here is that workers will *re-die* after the SIGHUP and the new children will not have the failed stream.

Pchelolo: @d00rman And it's actually what we want - if log rotation freed up some space, we want to begin logging again, don't we? And if the space wasn't freed up, we'd catch this error again and remove the file logger again..

Also, if we'd manually freed up some space, we'd what to do a rolling restart to recreate the loggers back

Pchelolo: > I guess the worst-case scenario here is that workers will re-die after the SIGHUP and the new children will not have the failed stream.

Re-die? Workers don't die because of this any more. When the partition is filled up the error is caught and the appropriate logger is removed. Workers continue to operate normally.

lib/logger.js
Pchelolo: A Logger with no streams is kinda a `null` logger already, isn't it?

d00rman: Duh, good point

Pchelolo:
Yay!

d00rman:
Published as [v2.2.3](https://github.com/wikimedia/service-runner/tree/v2.2.3)

lib/logger.js
```diff
@@ -121,6 +121,29 @@ var streamConverter = {
 };
 var streamConverterList = Object.keys(streamConverter);
 
+Logger.prototype._addErrorHandler = function() {
+    var self = this;
+    self._logger.on('error', function(err, failedStream) {
+        // If some fatal error occurred in one of the logger streams,
+        // we can't do much, so ignore.
+
+        if (err.code === 'ENOSPC') {
```
gwicke: Is there a reason to keep going if the error is not ENOSPC? For example, a partition that's remounted read-only should not take out the service.

Pchelolo: @gwicke I was thinking about that too. On one hand, we don't wanna stop logging if the error is recoverable, like some intermittent write failure. On the other hand, we wanna get rid of the logger for irrecoverable error.. I'm not sure where in the spectrum we wanna be...

gwicke: We normally still log to logstash, so imho we should err on the side of caution for local logging. There is no way for us to distinguish temporary failures from permanent ones, so I think we should just stop logging on any error. In any case, write failures are an exceptional situation, so this shouldn't affect normal operation.

Pchelolo: @gwicke Hm.. I can remove that `if` statement. I guess that should be true for any stream type - if the stream has failed, remove it and just log the failure using other streams. Sounds good?

gwicke: @Pchelolo: For remote UDP streams temporary failures are more likely, and we know that those don't cause memory issues. Lets perhaps limit disabling streams to file & stdout streams, or anything-but-gelf?

PR #158, opened 2017-03-03
Implemented sampled rate-limiting for logging

Pchelolo:
We can't reuse the existing rate_limiter, because we want a bit different thing - when the limit is about to be reached we want to probabilistically decide whether to log or not, so that we don't cut off all logging completely when the limit is reached, but instead start automatically sampling.

Bug: https://phabricator.wikimedia.org/T159479
cc @wikimedia/services

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,12 @@ logging:
   # regardless of the configured logging level.
   sampled_levels:
     'trace/webrequest': 0.2
+  # Set up the limiter for logging. When the rate of logs is close to the limit,
+  # it starts to increase the probability of discarding the logs of same kind.
+  # For default parameters use rate_limit: true
+  rate_limit:
+    limit: 2000 # How many per interval to log. Default: 1000
+    interval: 100 # Interval of rate sampling, ms. Default: 1000    
   streams:
   # Use gelf-stream -> logstash
   - type: gelf
--- a/lib/log_sampler.js
+++ b/lib/log_sampler.js
@@ -19,7 +19,7 @@ function Sampler(options) {
     this._counters = {};
     // Exponential decay with factor 2, thrice per interval for better
     // smoothness, using the cubic root of 2).
-    this._decayFactor = Math.pow(2, 1/3);
+    this._decayFactor = Math.pow(2, 1 / 3);
     this._decayedLimit = this._options.limit / this._decayFactor;
     this._decayInterval = setInterval(this._decay.bind(this), this._options.interval / 3);
 }
--- a/lib/log_sampler.js
+++ b/lib/log_sampler.js
@@ -1,0 +1,59 @@
+"use strict";
+
+/**
+ * Creates a log sampler. In case the log rate is not exceeded, all the logs
+ * are being written. Otherwise starts sampling the logs with decreasing
+ * probability.
+ *
+ * @param {object} options:
+ * @param {Number} [options.limit] The maximum rate of sampled logs per interval. Default: 1000.
+ * @param {Number} [options.interval] Update interval in ms. Default: 1000ms.
+ * @param {Number} [options.minValue] Drop global counters below this value. Default: 10.
+ */
+function Sampler(options) {
+    this._options = options === true ? {} : options || {};
+    this._options.limit = this._options.limit || 1000;
+    this._options.interval = this._options.interval || 1000;
+    this._options.minValue = this._options.minValue || 10;
+    // TODO: replace with a native Map when support for node 0.10 is dropped.
+    this._counters = {};
+    // Exponential decay with factor 2, thrice per interval for better
+    // smoothness, using the cubic root of 2).
+    this._decayFactor = Math.pow(2, 1/3);
+    this._decayedLimit = this._options.limit / this._decayFactor;
+    this._decayInterval = setInterval(this._decay.bind(this), this._options.interval / 3);
+}
+
+Sampler.prototype._decay = function() {
+    var self = this;
+    var minValue = self._options.minValue;
+    Object.keys(self._counters).forEach(function(key) {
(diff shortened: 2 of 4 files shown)


lib/log_sampler.js
```diff
@@ -0,0 +1,59 @@
...
+Sampler.prototype.shouldLog = function(logClass) {
+    this._increment(logClass);
+    var currentCount = this._counters[logClass];
+    if (currentCount < this._decayedLimit) {
+        // If it's not close to the limit yet - don't sample.
+        return true;
+    } else {
+        // We decay by 2 each interval, so the counter is roughly 1.5 time larger
+        // then continuous rate.
+        return Math.random() < 1.5 * this._decayedLimit / currentCount;
```
gwicke: For an even sampling behavior, you probably want to calculate the sampling rate on each decay trigger. As a side effect, this should also be slightly faster.

lib/log_sampler.js
```diff
@@ -0,0 +1,59 @@
...
+    this._options.limit = this._options.limit || 1000;
+    this._options.interval = this._options.interval || 1000;
+    this._options.minValue = this._options.minValue || 10;
+    // TODO: replace with a native Map when support for node 0.10 is dropped.
+    this._counters = {};
+    // Exponential decay with factor 2, thrice per interval for better
+    // smoothness, using the cubic root of 2).
+    this._decayFactor = Math.pow(2, 1 / 3);
+    this._decayedLimit = this._options.limit / this._decayFactor;
+    this._decayInterval = setInterval(this._decay.bind(this), this._options.interval / 3);
```
gwicke: If you calculate the sampling rate right after decay, then you don't need to schedule extra updates, as the only time the rate will change would be once per interval.

lib/log_sampler.js
Pchelolo: @gwicke But then it wouldn't react to very fast burst of logs. If start logging with insane rate, the sampling rate wouldn't change until the next decay, passing through all the crazy amount of logs..

README.md
```diff
@@ -130,6 +130,12 @@ logging:
   # regardless of the configured logging level.
   sampled_levels:
     'trace/webrequest': 0.2
+  # Set up the limiter for logging. When the rate of logs is close to the limit,
+  # it starts to increase the probability of discarding the logs of same kind.
+  # For default parameters use rate_limit: true
+  rate_limit:
```
gwicke: An alternative way of exposing this would be as a replacement of sampled_levels, perhaps as "level_rates", with a maximum rate for each level.

lib/log_sampler.js
gwicke: That depends on the interval. A burst over 1s does not seem to be that likely to cause big issues.

README.md
Pchelolo: Hm.. The `sampled_levels` is a bit different - there we want to log something not legged normally with some small rate, while here we do not want to log something that's normally logged with a crazy rate. So conceptually it's pretty different..

lib/log_sampler.js
Pchelolo: @gwicke True, but I think it's better to have a smooth stream of logs then 1s burst followed by nothing (the counter will get very high during the 1s burst) followed by another burst etc..

README.md
gwicke: Both are rather similar in that we are interested in getting a sampled set of data points, with a limited overall rate. The main downside I see in fully replacing static sampling rates with dynamic ones is that it will be slightly harder to reason about the resulting distribution of data points if the underlying event is very bursty.

lib/log_sampler.js
gwicke: If the event is produced at a steady rate, there will be a steady sample rate after the first interval. This is true in either scheme; if you want more frequent updates, you can just set the interval lower.

Pchelolo: "after the first interval" is the key here - if it's, let's say, a short-term problem connecting to Cassandra, there will be a very short spike in logs (up to hundreds thousands pre second with a very low duration). Then we won't sample this initial spike and that's exactly what we want to do. So basically we would miss the whole point of this feature. I don't think saving 1 floating point division per log call worths it.

gwicke: For a constant event stream, the current code will create a sawtooth profile, with sampling rates increasing after each interval, then dropping as the counter is incremented. IMHO, a steady sample rate would be more desirable in that case.

Have short logging bursts actually caused any issues in practice? I would think that UDP would make sure that excess logging would be dropped transparently.

README.md
gwicke: You are right though that the matching semantics are different. The normal level selector semantics are "this level and up", while sampled_levels has no such implication, so can be used to target specific lower-level events.

I was wondering though if it would make sense to unify them under the hood, and perhaps even at a higher level. One (strawman) way of doing so could perhaps be something like this:

```yaml
level_rates: 
  >=warn: 200
  trace/webrequest: 10
```

This would set a target rate limit of 200/s for warnings & higher, and 10/s for trace/webrequest.

Pchelolo: What would the semantics be if we say
```yaml
level_rates:
  >=warn: 200
  >=error: 400
```

Also `>=warn` is not a legal `yaml` key..

gwicke: @Pchelolo, in the case of multiple ranges we could choose one of these semantics:

- Fancy: Apply them in level order. In this case, `error` and `fatal` could log 400/s, while `warn` would be limited at 200.
- Simple: Only allow one range, and throw an error if more were supplied.

Re keys, with quoting `'>=warn'` is perfectly legal.

gwicke:
We lowered the priority of the corresponding task to "low": https://phabricator.wikimedia.org/T139674

This means that this is on the backburner for now.

PR #160, opened 2017-03-07
Docker: Ignore Docker version check errors

d00rman:
Docker has recently switched to a new versioning scheme modelled after the Ubuntu one - YY.MM.BB - with zero-padded fields. Alas, semver considers such versions invalid and throws an error. Hence, if it does that, assume that the Docker installed on the system is recent enough to proceed.

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -331,10 +331,16 @@ function ensureDockerVersion() {
         }
 
         var minimumDockerVersion = os.type() === 'Darwin' ? '1.12.0' : '1.8.0';
-        if (semver.lt(dockerVersion, minimumDockerVersion)) {
-            console.error('Building the deploy repo on ' + os.type() +
-                ' supported only with docker ' + minimumDockerVersion + '+');
-            process.exit(1);
+        try {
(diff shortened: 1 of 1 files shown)


d00rman:
cc @wikimedia/services

lib/docker.js
```diff
@@ -331,10 +331,16 @@ function ensureDockerVersion() {
         }
 
         var minimumDockerVersion = os.type() === 'Darwin' ? '1.12.0' : '1.8.0';
-        if (semver.lt(dockerVersion, minimumDockerVersion)) {
-            console.error('Building the deploy repo on ' + os.type() +
-                ' supported only with docker ' + minimumDockerVersion + '+');
-            process.exit(1);
+        try {
+            if (semver.lt(dockerVersion, minimumDockerVersion)) {
```
Pchelolo: Perhaps better `dockerVersion = dockerVersion.replace(/\.0+/g, '.')` to get rid of the 0-padding but leave the check in place?

d00rman: Muy bien @Pchelolo ! Will do so

d00rman:
Will open another PR with the suggested idea from @Pchelolo

PR #161, opened 2017-03-07
Docker: Adapt version checking to the new Docker versioning scheme

d00rman:
Docker has switched to Ubuntu-style versioning and now uses the format YY.MM.BB. The zero-padded components do not play well with semver's version checking, so remove them so as not to confuse semver.


Pchelolo:
LGTM

PR #162, opened 2017-03-07
Docker: Do not swallow all zeros when version-checking

d00rman:


d00rman:
cc @wikimedia/services

Pchelolo:
Oh, right...

PR #163, opened 2017-03-09
Upgrade to ES6 syntax

Pchelolo:
Parsoid is the last service still maintaining compatibility with node 0.10 and 0.12 and support period ends in April, so we can gradually update service-runner to ES6 syntax.

I't not using any node 6 features, so compatible with node 4.

All services tested locally with this version, no issues found, but I still want the rollout to be as gradual as possible, starting from less important like trending service.

One controversial thing here: we have the rule in general to limit the line length by 100 characters, but in `docker.js` using string templates some of the lines actually seem more readable to me not being split, so I've disabled that rule. If you think it's better to enable it back, I will.

This PR is made against a non-master branch, so we can merge, release and test this even before April.

Bug: https://phabricator.wikimedia.org/T160000 (Ha, look how round is the bug number!)
cc @wikimedia/services


Pchelolo:
@d00rman That's why I've created this PR against the `es6-upgrade` brunch and bumped the version to 3.0.0. That way we can support 2 parallel release branches: if something needs updating in current code we just push to master and release a new patch or server-minor patch. 

Having the 3.0.0 released allows us to get it out very slowly, service-by-service. Even if we wait for Parsoid to drop node 0.10, we'd probably do the same thing with 2 branches and 2 release lines, since during the upgrade we night find issues in non-upgraded services and not want to rush an upgrade to them.

Pchelolo:
It's April now, so we can move on this one.

Pchelolo:
Heeeeya @d00rman @gwicke let's go with this?

d00rman:
The announcement just got out. Let's do it! @Pchelolo have you rebased on top of current master? Maybe we can switch the PR to merge into master directly?

Pchelolo:
@d00rman Rebased, fixed ESLint etc

lib/docker.js
```diff
@@ -456,19 +422,15 @@ function updateDeploy() {
...
         }
-    }).then(function() {
-        // make sure the package.json symlink is in place
-        return fs.symlinkAsync(opts.submodule + '/package.json', 'package.json')
-        .catch(function() {}).then(function() {
-            return promisedGit(['add', 'package.json']);
-        });
-    }).then(function() {
+    }).then(() => // make sure the package.json symlink is in place
+    fs.symlinkAsync(`${opts.submodule}/package.json`, 'package.json')
```
d00rman: Indent

lib/docker.js
```diff
@@ -507,17 +468,13 @@ function updateDeploy() {
...
-            return promisedGit(['add', 'node_modules']);
-        });
-    }).then(function() {
-        // commit the changes
-        return promisedGit(['commit', '-m', opts.commit_msg]);
-    }).then(function() {
+        }).then(() => // add the built submodules
+        promisedGit(['add', 'node_modules']));
+    }).then(() => // commit the changes
+    promisedGit(['commit', '-m', opts.commit_msg])).then(() => {
```
d00rman: Idem.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "service-runner",
-  "version": "2.2.5",
+  "version": "3.0.0",
```
d00rman: Perhaps better to bump the version in a separate commit as it will give us better visibility of the switch to version 3

service-runner.js
```diff
@@ -22,34 +24,35 @@ if (cluster.isMaster) {
...
+    }
+
+    // @deprecated
+    run(conf) {
+        return this.start(conf)
+        .tap(() => {
+            // Delay the log call until the logger is actually set up.
+            if (this._impl._logger) {
+                this._impl._logger.log('warn/service-runner',
+                    'ServiceRunner.run() is deprecated, and will be removed in v3.x.');
```
d00rman: This will be ver 3.x, so perhaps remove this function altogether?

d00rman:
LGTM overall, just some minor nits here and there :+1:

Pchelolo:
@d00rman Updated. Reverted the version bump, will remove the deprecated function with the next PR with the version bump.

service-runner.js
Pchelolo: Will remove together with actually bumping the version of the package.

PR #166, opened 2017-10-31
Expose logger / metrics configs to workers

arlolra:
I'd like to make use of service-runner's metrics and logging libraries in subprocesses spawned from workers initiated by service-runner, and therefore need to expose configurations to them.

For example, see https://gerrit.wikimedia.org/r/#/c/355861/9/lib/config/ParsoidConfig.js

Any thoughts on an approach?

/cc @wikimedia/services @wikimedia/parsing

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -138,7 +138,9 @@ class Worker extends BaseService {
             } else {
                 basePath = this._basePath;
             }
-            service.conf.worker_id = this.config.worker_id;
+            ['worker_id', 'logging', 'metrics', 'num_workers'].forEach((k) => {
+                service.conf[k] = this.config[k];
+            });
             const opts = {
                 name,
                 appBasePath: basePath,
--- a/service-runner.js
+++ b/service-runner.js
@@ -12,6 +12,8 @@
 const cluster = require('cluster');
 const Master = require('./lib/master');
 const Worker = require('./lib/worker');
+const Logger = require('./lib/logger');
+const makeStatsD = require('./lib/statsd');
 
 // Disable cluster RR balancing; direct socket sharing has better throughput /
 // lower overhead. Also bump up somaxconn with this command:
@@ -52,6 +54,9 @@ class ServiceRunner {
             }
         });
     }
+
+    static getLogger(loggerConf) { return new Logger(loggerConf); }
+    static getMetrics(metricsConf, logger) { return makeStatsD(metricsConf, logger); }
 }
 
 module.exports = ServiceRunner;
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "service-runner",
-  "version": "2.3.0",
+  "version": "2.4.0",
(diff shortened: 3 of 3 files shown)


lib/worker.js
```diff
@@ -139,6 +139,10 @@ class Worker extends BaseService {
                 basePath = this._basePath;
             }
             service.conf.worker_id = this.config.worker_id;
+            service.conf.parent = {
+                logging: this.config.logging,
+                metrics: this.config.metrics,
+            };
```
Pchelolo: WE already expose the config to the user, maybe it's better to add these properties into the config?

Pchelolo:
The gerrit patch you've posted looks a little bit hacky to me as it goes into the internals of the service-runner library using knowledge on how the lib is organized. What I would propose instead is to make a public API in the ServiceRunner class to get logger and a metric reporter. It could be 2 static methods where you provide the configuration externally like `ServiceRunnner.getLogger(loggerConf)`.

Alternatively, we could make the methods non-static, so that the child processes would construct a ServiceRunner instance, but not start it and get a logger/metrix reported from it and the ServiceRunner instance would be responsible for loading the config from a standard location.

lib/worker.js
arlolra: I'm not sure what you mean.  Isn't that what this is doing?

arlolra:
> What I would propose instead is to make a public API in the ServiceRunner class to get logger and a metric reporter. It could be 2 static methods where you provide the configuration externally like ServiceRunnner.getLogger(loggerConf).

Ok, I'll give that a try.

arlolra:
@Pchelolo Please see the update.

Also, the gerrit patch with this change, https://gerrit.wikimedia.org/r/#/c/355861/10/lib/config/ParsoidConfig.js

lib/worker.js
Pchelolo: Oh, indeed, sorry. 

Not sure about the `parent` though. Do you think it's needed to wrap them into `parent`?

service-runner.js
```diff
@@ -54,6 +56,9 @@ class ServiceRunner {
     }
 }
 
+ServiceRunner.getLogger = (loggerConf) => { return new Logger(loggerConf); };
```
Pchelolo: I'd prefer method-like syntax within the class

```
static getLogger(loggerConf) { return new Logger(loggerConf); }

lib/worker.js
arlolra: There's already a key at the top level called `metrics`.  I'm not married to `parent` though, suggestions welcome.

service-runner.js
arlolra: Oh, sure, I was unfamiliar with that syntax.  Will fix.

arlolra:
@Pchelolo Ok, should be good to go.

Pchelolo:
Published as 2.4.0.

arlolra:
Thanks!

PR #167, opened 2017-11-06
Proper graceful shutdow

Pchelolo:
mocha 4.0 would stop killing the process after running the tests, so we need to implement proper graceful server shutdown - so clean all the intervals/timeouts.

Also the test server will need access to the `Server` object, so when we're running in `num_workers === 0` mode we don't really send the serverRet object to another process, so we don't need to check if it's serializable.

cc @wikimedia/services


d00rman:
Published as [v2.4.1](https://github.com/wikimedia/service-runner/tree/v2.4.1)

PR #168, opened 2017-11-07
Cascade close all child metrics reportes

Pchelolo:
We need to cascade-close all the children of a certain metrics reported on exit

cc @wikimedia/services


d00rman:
Published as [v2.4.2](https://github.com/wikimedia/service-runner/tree/v2.4.2

PR #170, opened 2017-11-21
Report GC statistics to statsd; rel v2.4.3

Pchelolo:
It would be interesting to monitor not only service heap, but service GC pauses as well. Right now we're experiencing some perf issues regarding change-prop, but this metric could be interesting generically for all services.

However, this brings in a binary dependency to service-runner. We install `heapdump` by default anyway, so I guess it's not a big deal, but we only actually install `heapdump` in a docker file for production. Maybe better to have the same approach for the gc metrics reporter as well? Bikeshed opportunity.

cc @wikimedia/services

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -221,7 +221,7 @@ function createDockerFile() {
         if (pkg.deploy.install_opts) {
             installOpts += `${pkg.deploy.install_opts.join(' ')} `;
         }
-        contents += `CMD ${beforeInstall}${npmCommand} install${installOpts}${afterInstall} && npm install heapdump `;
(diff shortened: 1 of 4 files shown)


d00rman:
I agree 100% that such stats would be valuable for diagnosing problems in production environments and I think we should have them. That said, depending on a binary module reduces service-runner's versatility and increases the chances of deploys going wrong (if the build script is not used), so I'd prefer not to have it as a hard dependency. I would vote for GC stats being collected only if the package is present on the system (i.e. the user explicitly installed it). For that, it would be good to add it to the install procedure in `lib/docker.js`.

Pchelolo:
@d00rman Update the PR to only install `gc-stats` in docker build and ignore reporting if it's not available.

d00rman:
Published as [v2.4.3](https://github.com/wikimedia/service-runner/tree/v2.4.3)

PR #171, opened 2017-11-27
Make GC metrics reported each second

Pchelolo:
The longer the workers live the more often they begin doing GC which floods metrics. Make the GC metrics go with a steady flow.

cc @wikimedia/services

### DIFF
--- a/lib/heapwatch.js
+++ b/lib/heapwatch.js
@@ -2,6 +2,14 @@
 
 const cluster = require('cluster');
 
+const ZERO_CUMULATIVE_GC_INTERVAL = {
+    minor: 0,
+    major: 0,
+    incremental: 0,
+    weak: 0
+};
+const GC_REPORT_INTERVAL = 1000;
+
 class HeapWatch {
     constructor(conf, logger, statsd) {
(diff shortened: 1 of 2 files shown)


lib/heapwatch.js
```diff
@@ -30,7 +40,19 @@ class HeapWatch {
                 // Report memory stats to statsd. Use 'timing' (which isn't really
                 // timing-specific at all) instead of 'gauge' to get percentiles &
                 // min/max.
-                this.statsd.timing(`gc.${gcTypeName(stats.gctype)}`, stats.pause);
+                const type = gcTypeName(stats.gctype);
+                if (type !== 'unknown') {
+                    this.cumulativeGCTimes[gcTypeName(stats.gctype)] += stats.pause;
+                }
+                this.gcReportInterval = setInterval(() => {
```
d00rman: The reporting interval should be set only once, not every time the event fires.

Pchelolo: Done

d00rman:
Published as [v2.4.4](https://github.com/wikimedia/service-runner/tree/v2.4.4)

PR #172, opened 2017-11-27
Use Object.assign to clone the constant zeros

Pchelolo:
Obvious bug - we were changing the const object that was supposed to be unchanged.

cc @wikimedia/services


d00rman:
Published as [v2.4.5](https://github.com/wikimedia/service-runner/tree/v2.4.5)

PR #173, opened 2017-11-27
GC: Send the stats only once they have been cleared; rel v2.4.6

d00rman:
Sending stats requires network I/O, which means that the GC `stats` event could fire while we are sending them, which pollutes the statistics. In order to fix that, first only collect the stats to send, then clear them and finally send them.


Pchelolo:
Published as `v2.4.6`

PR #175, opened 2017-12-04
Only set GC monitor once

Pchelolo:
I will just not write anything here and let us enjoy how idiotic the bug was

cc @wikimedia/services


Pchelolo:
I will self-merge and deploy because this might actually affect metrics collection.

PR #176, opened 2017-12-06
Fix: Always install npm; release v2.4.8

d00rman:
The `npm` system package depends on various other nodejs-related packages that are sometimes needed to build binary dependencies, so make sure that it is installed regardless of the actual npm version used.


d00rman:
Published as [v2.4.8](https://github.com/wikimedia/service-runner/tree/v2.4.8)

PR #178, opened 2018-01-19
Allow using custom NPM version in the build

Pchelolo:
Sometimes npm shipped with node has bugs that break the build
for certain packages. To be able to quickly workaround such bugs
we allow to select a custom npm version that should be used.

Also, the reshrinkwrap setting was never used since it's creation
and is not really needed in the future with introduction of
packages-lock.json file in npm v4, so it was removed.

cc @wikimedia/services

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,6 @@ Options:
                                              [string] [default: "./config.yaml"]
   -f, --force        force the operation to execute   [boolean] [default: false]
   -d, --deploy-repo  build only the deploy repo       [boolean] [default: false]
-  -s, --reshrinkwrap re-generate shrinkwrap.json      [boolean] [default: false]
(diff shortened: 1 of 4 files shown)


d00rman:
Hm, this installs npm locally, which means npm and its dependencies would end up in production. Would it be possible to install the specific npm version globally rather than locally?

Pchelolo:
@d00rman there's a piece that deletes the installed local npm after the dependencies installation is complete, so npm doesn't end up in production.

It might be possible to install npm globally with a `-g` flag, but in order to do that, we'd need to run the command inside a container as root, which we were trying to avoid.

Pchelolo:
@d00rman done, done and done.

d00rman:
Published as [v2.5.0](https://github.com/wikimedia/service-runner/tree/v2.5.0)

PR #179, opened 2018-01-19
Do not exit right away if there was a log message before.

Pchelolo:
Logging is async, so we need to allow some delay before calling
process.exit in order to let the logger process the message.

Follow-up for #177 

cc @wikimedia/services @arlolra


arlolra:
Thanks, LGTM.

PR #180, opened 2018-02-08
Normalize the logging messages for easier filtering in kibana

Pchelolo:
The additional information has been moved to separate fields in logs,
like this we can easily filter out mesages about worker deaths and
filter a sequence of events that lead to the death of a particular
worker.

Bug: https://phabricator.wikimedia.org/T186761
cc @wikimedia/services @berndsi @mdholloway

### DIFF
--- a/lib/master.js
+++ b/lib/master.js
@@ -98,7 +98,7 @@ class Master extends BaseService {
         return this._ratelimiter.setup()
         .then(() => this._startWorkers(this.config.num_workers))
         .tap(() => {
-            this._logger.log('warn/service-runner', 'Startup finished');
+            this._logger.log('warn/service-runner', 'startup finished');
             this._checkHeartbeat();
         });
     }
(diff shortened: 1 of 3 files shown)


lib/master.js
```diff
@@ -175,26 +175,28 @@ class Master extends BaseService {
...
-                            // Give up.
-                            this._logger.log('fatal/service-runner/master',
-                                'startup failed, exiting master');
-                            return this._exitProcess(1);
-                        }
+                    if (!this._firstWorkerStarted
+                            && this._firstWorkerStartupAttempts++ >= STARTUP_ATTEMPTS_LIMIT) {
+                        // We tried to start the first worker 3 times, but never succeed.
+                        // Give up.
+                        this._logger.log('fatal/service-runner/master',
```
d00rman: Would be good to have the exit code

d00rman:
LGTM modulo small nit

lib/master.js
Pchelolo: @d00rman done

Pchelolo:
Tagged and published as service-runner@2.5.1

d00rman:
That was fast!

berndsi:
@Pchelolo I think you wanted to use `msg` instead of `message`. I'm getting duplicate entries:

{"name":"service-mobileapp-node","hostname":"box","pid":6131,"level":30,"**message**":"worker shutting down","worker_pid":6131,"levelPath":"info/service-runner/worker","**msg**":"worker shutting down","time":"2018-02-08T16:36:24.717Z","v":0}

Pchelolo:
Oh holy hell I thought we've fixed the msg/message thing. Will fix.

On the `worker_pid` thing - this particular log entry is from the worker, so yes, the worker pid will match with pid. but In case you want to filter logs to know the whole sequence of logs coming from a specific worker - including the `worker_pid` is helpful cause some of the messages will be coming from the master where the `pid` is different.

berndsi:
Then it's ok to keep the `worker_pid`. Thanks!

PR #181, opened 2018-02-14
Don't install npm for build in case nvm will be used to install it.

Pchelolo:
The npm package is not available in Debian Stretch, so the build
that's using it as a base image fails. However, we don't even use npm
installed from the apt repo as most of the time we use nvm to install.

See: https://github.com/nodejs/help/issues/1040
cc @wikimedia/services @niedzielski

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -110,7 +110,23 @@ function promisedSpawn(args, options) {
  */
 function createDockerFile() {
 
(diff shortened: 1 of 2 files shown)


d00rman:
Published as [v2.5.2](https://github.com/wikimedia/service-runner/tree/v2.5.2)

PR #183, opened 2018-03-25
Fix Docker working directory permissions

mdholloway:
The current Dockerfile leaves root as the owner of the working directory,
which can cause permissions issues in the event that the run user created
on non-Mac platforms needs to modify files in the working directory.
Running docker-test for mobileapps on Linux currently fails for this
reason.  This patch adds a command to change ownership of the working
directory to runuser:rungroup.

### DIFF
--- a/lib/docker.js
+++ b/lib/docker.js
@@ -219,7 +219,9 @@ function createDockerFile() {
             // In 'Docker for Mac' the mapping between users/groups
             // is done internally by docker, so we can run as root
             && os.type() !== 'Darwin') {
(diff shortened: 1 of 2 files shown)


Pchelolo:
Hm.... In production the services run as root inside firejail so I'm not 100% that this wouldn't break the production...

mdholloway:
AIUI this change only concerns the Docker container built for the deploy repo build process, which is destroyed after that process completes, so I don't think this change should be relevant to production.  That might change if we begin running services in containers in production via docker-start (perhaps with the planned move to Kubernetes) but AIUI that's not what's happening now.

d00rman:
The main reason the directory is owned by root is precisely *because* we don't want the run user to be able to modify the files there, as that presents a serious security concern.

mdholloway:
Point taken; and also docker-test now works in a newly created Ubuntu VM (I removed the old disk partitions on which I was running into this so at this point I can no longer repro).  I'll close this.

PR #184, opened 2018-04-17
Support concurrent worker startup

Pchelolo:
Without adding the `startup_concurrency` config variable this doesn't change the current behaviour.

cc @wikimedia/services

### DIFF
--- a/lib/master.js
+++ b/lib/master.js
@@ -108,15 +108,18 @@ class Master extends BaseService {
             message: 'SIGHUP received, performing rolling restart of workers'
         });
         this._inRollingRestart = true;
-        P.each(Object.keys(cluster.workers), workerId => this._stopWorker(workerId)
-        .then(() => this._startWorkers(1)))
+        P.each(Object.keys(cluster.workers), (workerPid) => {
+            const workerId = cluster.workers[workerPid].worker_id;
+            return this._stopWorker(workerPid)
+            .then(() => this._startWorker(workerId));
+        })
         .then(() => {
             this._inRollingRestart = false;
         });
     }
 
-    _stopWorker(workerId) {
-        const worker = cluster.workers[workerId];
+    _stopWorker(workerPid) {
+        const worker = cluster.workers[workerPid];
         if (!worker || worker.state === 'disconnected') {
             if (worker) {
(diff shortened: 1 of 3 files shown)


lib/master.js
```diff
@@ -153,112 +156,108 @@ class Master extends BaseService {
...
-            return res;
-        }
+        });
+    }
+    // Fork a single worker, wait for it to start executing and set everything up,
+    // and then fork all the rest of the workers.
+    _startWorkers(workersToStart) {
+        return this._startWorker(1)
+        .then(firstReturn => P.map(
+            Array.from(new Array(workersToStart - 1), (val, i) => i + 2),
```
d00rman: I'm not entirely clear why (i) you are using the `(val, i)` form of the mapper function; and (ii) why you skip one index with `i + 2`

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "service-runner",
-  "version": "2.5.2",
-  "description": "Generic nodejs service supervisor / cluster runner",
+  "version": "2.6.0",
+  "description": "G eneric nodejs service supervisor / cluster runner",
```
d00rman: Oops

lib/master.js
Pchelolo: (i) `new Array(workersToStart - 1)` with produce a `workersToStart - 1` element array of `undefined`, so to convert it to array of `2,3,4...` we need an index, the `val` will always be `undefined`

(ii) These indexes are used as the worker ID. eg, worker number 1, number 2 etc. WE're starting the first worker (number 1) sequentially, so here we want to generate an array of remaining IDs, eg [2,3,4,...,num_workers].

Pchelolo:
> I tested it and everything works as advertised, except a minor thing: when startup_concurrency > 1, issuing a SIGHUP starts all the workers serially instead of starting them in batches. Let's fix that and then we are good to go.

That actually was intentional. The method is called `rollingRestart`, so I figured that we need to actually rolling restart them one-by-one. Happy to discuss this though.

d00rman:
Published as [v2.6.0](https://github.com/wikimedia/service-runner/tree/v2.6.0)

PR #186, opened 2018-05-01
Get rid of extend dependency and protect the log info object.

Pchelolo:
When logging an error like `log('trace/error', err)`
we should not pass the actual instance of teh error object
into bunyan, because it assigns a bunch of properties to
the object the we do not want to publicly expose, like
_rootReq that has a `x-client-ip` header. Also it when
serializing it it apparently messes up with an original object,
so later on the original error gets wrapped into a 500.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const extend = require('extend');
 const bunyan = require('bunyan');
 const gelfStream = require('gelf-stream');
 const syslogStream = require('bunyan-syslog-udp');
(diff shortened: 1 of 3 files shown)


lib/logger.js
```diff
@@ -24,7 +23,7 @@ const streamConverter = {
             const selectedPort = parseInt(selectedHost.substr(selectedHost.indexOf(':') + 1), 10);
 
             if (selectedHost && !Number.isNaN(selectedPort)) {
-                host = selectedHost;
+                ho  pst = selectedHost;
```
d00rman: Euh?

Pchelolo: Fixed. Typo...

d00rman:
Published as [v2.6.1](https://github.com/wikimedia/service-runner/tree/v2.6.1)

PR #187, opened 2018-06-04
Preserve the Error type when logging errors.

Pchelolo:
We need to clone the info object when logging if it
is an Error since bynuan modifies the original object,
and that error is then reused to report the error to
the client. But we need to preserve the Error type so
that proper bynuan serializers kick in.

Bug: https://phabricator.wikimedia.org/T196124

cc @wikimedia/services @berndsi @mdholloway

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -307,7 +307,7 @@ class Logger {
 
             // Also pass in default parameters
             if (info instanceof Error) {
-                // We want to preserve Error type so that standard banyan serializers kick in.
+                // We want to preserve the Error type before bunyan serialisation kicks in.
                 info = Object.assign(copyError(info), this.args);
             } else {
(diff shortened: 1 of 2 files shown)


lib/logger.js
```diff
@@ -305,7 +306,12 @@ class Logger {
             const msg = this._createMessage(info);
 
             // Also pass in default parameters
-            info = Object.assign({}, info, this.args);
+            if (info instanceof Error) {
+                // We want to preserve Error type so that standard banyan serializers kick in.
```
d00rman: *We want to preserve the Error type before bunyan serialisation kicks in*

Pchelolo: Done.

d00rman:
Published as [v2.6.2](https://github.com/wikimedia/service-runner/tree/v2.6.2)

PR #188, opened 2018-06-05
More generically copy the errors

Pchelolo:
The library used for copying before was not suited for our usage of the `HTTPError`, so just use the object.assign and clone the prototype.

cc @wikimedia/services


d00rman:
Published as [v2.6.3](https://github.com/wikimedia/service-runner/tree/v2.6.3)

PR #189, opened 2018-06-13
Do not instantiate rate limiter if it's not required

Pchelolo:
In master-worker configuration the rate limiter sends quite a few messages back and forth from master to worker processes. If the rate limiter is not needed and not configured - just don't create it.

cc @wikimedia/services

### DIFF
--- a/lib/master.js
+++ b/lib/master.js
@@ -93,9 +93,12 @@ class Master extends BaseService {
             .then(this._rollingRestart.bind(this));
         });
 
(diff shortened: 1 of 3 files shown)


Pchelolo:
Hm.. I see what you're saying... The difference between the rate limiter and others is that the rate limiter is actually pretty expencive, so here I think it's better to opt-in for it then to opt-out

d00rman:
That's true, makes sense in this case.

d00rman:
Published as [v2.6.4](https://github.com/wikimedia/service-runner/tree/v2.6.4)

PR #191, opened 2018-11-08
Introduce optional promise cancellation feature

polishdeveloper:
Some projects require the Bluebird promise cancellation feature.
That feature has to be enabled before any promise is created.
Because ServiceRunner service relies havily on the promises, and
it runs promises before it starts executing any application code
the only place to put that config is the ServiceRunner itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,8 @@ var startupPromise = runner.start({
    you must override the app base path with either:
      - `APP_BASE_PATH` environment variable
      - `app_base_path` config stanza.
+- If project requires cancellable promises (which is disabled by default) you must set the
+`APP_ENABLE_CANCELLABLE_PROMISES` environment variable.
 - Default top-level config format (**draft**):
 
 ```yaml
--- a/README.md
+++ b/README.md
@@ -110,7 +110,8 @@ var startupPromise = runner.start({
      - `APP_BASE_PATH` environment variable
      - `app_base_path` config stanza.
 - If project requires cancellable promises (which is disabled by default) you must set the
-`APP_ENABLE_CANCELLABLE_PROMISES` environment variable.
+`APP_ENABLE_CANCELLABLE_PROMISES` environment variable to non-empty value (like `1` or `true`). For more information about
+ cancellable promises please refer to [Bluebird documentation](http://bluebirdjs.com/docs/api/cancellation.html).
 - Default top-level config format (**draft**):
 
 ```yaml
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -109,6 +109,8 @@ var startupPromise = runner.start({
    you must override the app base path with either:
      - `APP_BASE_PATH` environment variable
      - `app_base_path` config stanza.
+- If project requires cancellable promises (which is disabled by default) you must set the
```
Pchelolo: Could you add a link to the documentation regarding cancellable promises here?

README.md
```diff
@@ -109,6 +109,8 @@ var startupPromise = runner.start({
...
+`APP_ENABLE_CANCELLABLE_PROMISES` environment variable.
```
Pchelolo: Could you also specify the value the variable should be set to?

Pchelolo:
Minor nit picking, otherwise LGTM

Pchelolo:
This LGTM, but I'd like @d00rman to have a look on our judgement here as well. The fact it's not published on NPM doesn't block you I guess?

README.md
d00rman: `If the project requires cancellable promises (which are disabled by default)`

README.md
```diff
@@ -109,6 +109,9 @@ var startupPromise = runner.start({
    you must override the app base path with either:
      - `APP_BASE_PATH` environment variable
      - `app_base_path` config stanza.
+- If project requires cancellable promises (which is disabled by default) you must set the
+`APP_ENABLE_CANCELLABLE_PROMISES` environment variable to non-empty value (like `1` or `true`). For more information about
```
d00rman: `to a non-empty and truth-y value`

service-runner.js
```diff
@@ -62,6 +63,13 @@ class ServiceRunner {
 module.exports = ServiceRunner;
 
 if (module.parent === null) {
+    // Cancellable promises have to enabled before we start any promise. Because ServiceRunner
```
d00rman: `s/start any promise/instantiate any promises/`

service-runner.js
```diff
@@ -62,6 +63,13 @@ class ServiceRunner {
...
+    // heavily rely on promises this is the best place to leave the config like that
```
d00rman: `s/rely/relies/`

Pchelolo:
Published!

PR #193, opened 2018-12-17
Add the env var for the config file path; release v2.6.8

d00rman:
For WMF production needs, we need a more flexible way of declaring the location of the config file, so have `service-runner` check for the `APP_CONFIG_PATH` environment variable. If specified, it overrides any command-line options given.

Bug: [T212113](https://phabricator.wikimedia.org/T212113)

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,14 +104,20 @@ var startupPromise = runner.start({
 ### Config loading
 - Default config locations in a project: `config.yaml` for a customized config,
     and `config.example.yaml` for an example config for a service.
+- You can specify the location of the configuration file in two ways: by using
+  the `-c`/`--config` command-line option; or by setting the `APP_CONFIG_PATH`
+  environment variable. If both are specified, the environment variable takes
+  precedence.
  - By default, we assume that your project depends on `service-runner` and
    you follow standard node project layout. However, if a custom layout is used,
    you must override the app base path with either:
      - `APP_BASE_PATH` environment variable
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -104,14 +104,20 @@ var startupPromise = runner.start({
 ### Config loading
 - Default config locations in a project: `config.yaml` for a customized config,
     and `config.example.yaml` for an example config for a service.
+- You can specify the location of the configuration file in two ways: by using
+  the `-c`/`--config` command-line option; or by setting the `APP_CONFIG_PATH`
+  environment variable. If both are specified, the environment variable takes
+  precedence.
```
Pchelolo: I would argue that it should be vice-versa and the command line option should take precedence.

d00rman: I wouldn't agree @Pchelolo. The command-line argument is kind of more permanent than environment variables (they are "variable" by definition), which means that there is a greater chance of the env var changing than the cmd argument.

Pchelolo: On the other hand, in case several services are running in the same environment - there will be no way to set different paths for different services. To me, it feels like the env variable is the 'default' and you could override it with cmd argument

d00rman: I don't understand your argument. Env vars can be set on a per-process basis.

Pchelolo: ye but you also can set it globally and then there will be no way to override it. However, I'm nitpicking. overall +1 to the patch

PR #195, opened 2019-01-11
heapwatch: update GC timing stats comment

filippog:
cc @Pchelolo


Pchelolo:
Copy-paste artifact..

PR #196, opened 2019-02-22
Use hot-shots native childClient

hknustwmf:
Replaced statsd client instances creation with childClient method calls in makeChild. Upped the version number to 2.6.10.

Bug:  [T203617](https://phabricator.wikimedia.org/T203617)


lib/docker.js
```diff
@@ -212,7 +214,7 @@ function createDockerFile() {
     }
 
     if (!opts.deploy) {
-        contents += `RUN mkdir /opt/service\nADD . /opt/service\nWORKDIR /opt/service\nRUN ${npmCommand} install && npm dedupe\n`;
+        contents += `RUN mkdir /opt/service\nADD . /opt/service\nWORKDIR /opt/service\nRUN rm -rf node_modules && ${npmCommand} install && npm dedupe\n`;
```
d00rman: We don't do `rm -rf node_modules` here on purpose: we want developers to be able to quickly spin up their service

lib/docker.js
```diff
@@ -595,7 +600,7 @@ function main(options, configuration) {
         // set the defaults
         targets = {
             debian: 'debian:jessie',
-            ubuntu: 'ubuntu:14.04'
+            ubuntu: 'ubuntu:18.04'
```
d00rman: Let's not do changes that don't belong conceptually in this PR. Also, the last Ubuntu version we supported in prod was 14.04 so better to keep it there (even though we don't use Ubuntu in production any more at all).

hknustwmf: Reverted back

lib/docker.js
hknustwmf: Removed

Pchelolo:
first things first - let's not mix completely unrelated things into a single pull request. The docker.js changes are good, they just entirely do not belong here.

lib/statsd.js
```diff
@@ -95,8 +105,6 @@ function makeStatsD(options, logger) {
         if (typeof originalClose === 'function') {
             originalClose.apply(statsd);
         }
-        statsd._children.forEach(child => child.close());
-        statsd._children = [];
     })(statsd.close);
```
Pchelolo: We don't need this anymore at all. The whole redefinition of the 'close' method.

lib/statsd.js
```diff
@@ -71,15 +72,24 @@ function makeStatsD(options, logger) {
...
+        const child = statsd.childClient(childOptions);
+
+        // Attach normalizeName to make the childClient instance backwards-compatible
+        // with the previously used StatsD instance
+        child.normalizeName = normalizeName;
+
+        // Also add a small utility to send a delta given a startTime
+        child.endTiming = function endTiming(names, startTime, samplingInterval) {
+            return this.timing(names, Date.now() - startTime, samplingInterval);
+        };
```
Pchelolo: This code is duplicated on line 99. We can just make it a const endTiming and then bind it to an appropriate 'this'.

Alternatively event better, we can make a function that will apply the mixin to the original hot-shots statsd and share it for both the root and the child.

Pchelolo:
package-lock.js is here.

Pchelolo:
Did you test it with RB?

Pchelolo:
While you're on this repo, couble you make a separate PR to remove mocha-eslint and eslint-config-node-services? I forgot to add this repo into the list

lib/statsd.js
hknustwmf: Removed.

lib/statsd.js
hknustwmf: Added mixin

hknustwmf:
> Did you test it with RB?

Yes. I ran restbase to test the service-runner changes

Pchelolo:
The description of the PR is not correct - this one does not touch the Docker.js

d00rman:
Published as [v2.6.10](https://github.com/wikimedia/service-runner/tree/v2.6.10)

PR #198, opened 2019-02-26
Removed mocha-eslint and eslint-config-node-svc deps. 

hknustwmf:
Minor code reformatting to satisfy linter.


Pchelolo:
Neat, thank you. no need for a new release, it's a no-op

PR #199, opened 2019-03-13
Continue startup sequence if the worker dies during startup.

Pchelolo:
cc @wikimedia/services @ottomata


d00rman:
Published as [v2.6.11](https://github.com/wikimedia/service-runner/tree/v2.6.11)

PR #203, opened 2019-03-15
Changes to enable @CEE for T211125

hknustwmf:


Pchelolo:
oh, actually, could you also explicitly bump the syslog-udp dependency version

lib/logger.js
```diff
@@ -70,7 +70,8 @@ const streamConverter = {
         const impl = syslogStream.createBunyanStream({
             facility: stream.facility || 'local0',
             host: stream.host || '127.0.0.1',
-            port: stream.port || 514
+            port: stream.port || 514,
+            prefix: stream.prefix || ''
```
Pchelolo: The package also allows to set up the 'name'. Let's change this from explicit properties remapping to an `Object.assign` that would provide the defaults for the host, port and facility in order to allow additional configuration properties to be passed as is.

hknustwmf: Done

lib/logger.js
```diff
@@ -67,11 +67,14 @@ const streamConverter = {
...
-            port: stream.port || 514
-        });
+        const defaultOpts = {
+            facility: 'local0',
+            host: '127.0.0.1',
+            port: 514,
+            prefix: ''
+        };
+
+        const impl = syslogStream.createBunyanStream(Object.assign(stream, defaultOpts));
```
Pchelolo: This should be other way around. The right argument overrides the left argument.

PR #204, opened 2019-03-20
Provide sencible defaults for options passed via code

Pchelolo:
When the options are passed into the service-runner constructor via code (in tests for example) we need to provide the same defaults as for command-line arguments to ensure all the rest of the code works.


d00rman:
Published as [v2.6.13](https://github.com/wikimedia/service-runner/tree/v2.6.13)

PR #206, opened 2019-03-26
Provide a tests service runner

Pchelolo:


test/TestServer.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const P = require('bluebird');
+const ServiceRunner = require('service-runner');
```
d00rman: Did you mean(TM) `require('../service-runner.js')` ? As written, this might have unexpected consequences if, e.g., somebody has `service-runner` installed globally.

test/TestServer.js
```diff
@@ -0,0 +1,50 @@
...
+
+    start() {
+        if (this._running) {
+            console.log('The test server is already running. Skipping start.');
+            return P.resolve();
+        }
+
+        return this._runner.start()
+        .tap(() => {
+            this._running = true;
```
d00rman: Shouldn't this flag be set only if we actually managed to start the server? If the start fails, `.catch()` is activated, which calls the method again, but since the flag is set, it won't actually do anything, would it?

Pchelolo: `.tap` is just like `.then` so if it fails this will not be called.

test/TestServer.js
Pchelolo: Done

PR #208, opened 2019-03-29
Have gc-stats and heapdump as optionalDeps; release v2.6.17

d00rman:
Having `gc-stats` and `heapdump` being added to the build process post-facto is problematic for services being deployed to Kubernetes, as they won't be using the build script that comes with `service-runner`. Therefore, make them optional dependencies. This will allow us to have them installed in production after `npm install`, all the while not breaking the installation process for systems that do not support these modules.

Bug: [T205911](https://phabricator.wikimedia.org/T205911)


d00rman:
Published as [v2.6.17](https://github.com/wikimedia/service-runner/tree/v2.6.17)

PR #210, opened 2019-05-08
SIG* and event handler changes

hknustwmf:
The RESTBase tests run clean except for redirects which is still throwing "Warning: Possible EventEmitter memory leak detected. 11 pipe listeners added.". I assume this is actually caused by RESTBase itself rather than service-runner.

We probably want to bump the version number too.

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -39,7 +39,7 @@ class BaseService {
         this._inLogger = false;
 
         this._uncaughtExceptionHandler = (err) => {
-            this.logUnhandledException(err);
+            this._unhandledRejectionHandler(err);
             // Don't exit right away, allow logger to process message
             setTimeout(() => {
                 process.exit(1);
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -11,6 +11,7 @@ const Logger = require('./logger');
 const docker = require('./docker');
 
 const NUM_CPU_REGEX = /^(?:(?:ncpu[\s)*+-/])|[\s()*+-/.\d])+(?:ncpu)?$/;
+
 /**
  * Abstract base class for Master and Worker classes.
  *
@@ -32,13 +33,38 @@ class BaseService {
         this._logger = null;
         this._metrics = null;
         this._ratelimiter = null;
-        this.serviceReturns = null;
+        this._serviceReturns = null;
+
+        // Avoid recursion if there are bugs in the logging code.
+        this._inLogger = false;
+
+        this._uncaughtExceptionHandler = (err) => {
+            this.logUnhandledException(err);
+            // Don't exit right away, allow logger to process message
+            setTimeout(() => {
+                process.exit(1);
+            }, 100);
+        };
+
+        this._unhandledRejectionHandler = (err) => {
+            if (!this._inLogger) {
+                this._inLogger = true;
+                if (!err.stack) {
+                    Error.captureStackTrace(err);
+                }
+                this._logger.log('fatal/service-runner/unhandled', err);
+                this._inLogger = false;
+            }
+        };
     }
 
     stop() {
         if (this._ratelimiter) {
             this._ratelimiter.stop();
         }
+        // Remove handlers
+        process.removeListener('unhandledRejection', this._unhandledRejectionHandler);
+        process.removeListener('uncaughtException', this._uncaughtExceptionHandler);
     }
 
     _setAppBasePath(config) {
@@ -91,6 +117,12 @@ class BaseService {
 
             this._logger = new Logger(config.logging);
             this._setUpDNSCaching();
+
+            // Catch unhandled rejections & log them. This relies on bluebird.
+            process.on('unhandledRejection', this._unhandledRejectionHandler);
+
+            // Similarly, log uncaught exceptions. Also, exit.
+            process.on('uncaughtException', this._uncaughtExceptionHandler);
 
             return this._start();
         });
--- a/lib/heapwatch.js
+++ b/lib/heapwatch.js
@@ -21,28 +21,31 @@ class HeapWatch {
         this.timeoutHandle = undefined;
         this.gcReportInterval = undefined;
         this.cumulativeGCTimes = Object.assign({}, ZERO_CUMULATIVE_GC_INTERVAL);
+        this.reportStatsHandler = (stats) => {
+                // Report GC timings to statsd (in nanoseconds).
+                const type = this._gcTypeName(stats.gctype);
+                if (type !== 'unknown') {
+                    this.cumulativeGCTimes[this._gcTypeName(stats.gctype)] += stats.pause;
+                }
+        };
+        this._gcStats = null;
+    }
+
+    _gcTypeName(typeID) {
+        switch (typeID) {
+            case 1: return 'minor';
+            case 2: return 'major';
+            case 4: return 'incremental';
+            case 8: return 'weak';
+            case 15: return 'all';
+            default: return 'unknown';
+        }
     }
 
     setGCMonitor() {
         try {
-            const gcStats = require('gc-stats')();
-            const gcTypeName = (typeID) => {
-                switch (typeID) {
-                    case 1: return 'minor';
-                    case 2: return 'major';
-                    case 4: return 'incremental';
-                    case 8: return 'weak';
-                    case 15: return 'all';
-                    default: return 'unknown';
-                }
-            };
-            gcStats.on('stats', (stats) => {
-                // Report GC timings to statsd (in nanoseconds).
-                const type = gcTypeName(stats.gctype);
-                if (type !== 'unknown') {
-                    this.cumulativeGCTimes[gcTypeName(stats.gctype)] += stats.pause;
-                }
-            });
+            this._gcStats = require('gc-stats')();
+            this._gcStats.on('stats', this.reportStatsHandler);
             this.gcReportInterval = setInterval(() => {
                 const timings = {};
                 Object.keys(this.cumulativeGCTimes).forEach((gcType) => {
@@ -107,6 +110,9 @@ class HeapWatch {
     }
 
     close() {
+        if (this._gcStats) {
+            this._gcStats.removeListener('stats', this.reportStatsHandler);
+        }
         if (this.timeoutHandle) {
             clearTimeout(this.timeoutHandle);
             this.timeoutHandle = undefined;
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -97,7 +97,6 @@ class Logger {
             this._sampled_levels = conf.sampled_levels || {};
             delete conf.sampled_levels;
             this._logger = bunyan.createLogger(conf);
-            this._addErrorHandler();
             this._levelMatcher = this._levelToMatcher(conf.level);
             // For each specially logged component we need to create
             // a child logger that accepts everything regardless of the level
@@ -112,8 +111,27 @@ class Logger {
             this._traceLogger = this._logger.child({
                 level: bunyan.TRACE
             });
-            // Set up handlers for uncaught extensions
-            this._setupRootHandlers();
+
+            this._errorHandler = (err, failedStream) => {
+                // If some fatal error occurred in one of the logger streams,
+                // we can't do much, so ignore.
+
+                if (failedStream.type === 'file') {
+                    // However, if it's a `file` stream, it's likely that
+                    // we're in some catastrophic state. Remove the stream,
+                    // log the error and hope other streams would deliver it.
+                    // Attempting to continue logging through a failed `file`
+                    // stream might end up in a memory leak.
+                    this._logger.streams = this._logger.streams.filter((s) => s !== failedStream);
+                    failedStream.stream.destroy();
+                    // Hope that we have other streams to report the problem
+                    this.log('fatal/service-runner/logger', {
+                        message: 'Failed to write logs to file',
+                        error: err
+                    });
+                }
+            };
+            this._logger.on('error', this._errorHandler );
         } else {
             this._sampled_levels = confOrLogger._sampled_levels;
             this._logger = confOrLogger._logger;
@@ -121,29 +139,8 @@ class Logger {
             this._componentLoggers = confOrLogger._componentLoggers;
             this._traceLogger = confOrLogger._traceLogger;
         }
+
         this.args = args;
-    }
-
-    _addErrorHandler() {
-        this._logger.on('error', (err, failedStream) => {
-            // If some fatal error occurred in one of the logger streams,
-            // we can't do much, so ignore.
-
-            if (failedStream.type === 'file') {
-                // However, if it's a `file` stream, it's likely that
-                // we're in some catastrophic state. Remove the stream,
-                // log the error and hope other streams would deliver it.
-                // Attempting to continue logging through a failed `file`
-                // stream might end up in a memory leak.
-                this._logger.streams = this._logger.streams.filter((s) => s !== failedStream);
-                failedStream.stream.destroy();
-                // Hope that we have other streams to report the problem
-                this.log('fatal/service-runner/logger', {
-                    message: 'Failed to write logs to file',
-                    error: err
-                });
-            }
-        });
     }
 
     _processConf(conf) {
@@ -197,34 +194,6 @@ class Logger {
     _getLevelIndex(level) {
         const idx = LEVELS.indexOf(level);
         return idx !== -1 ? idx : DEF_LEVEL_INDEX;
-    }
-
-    _setupRootHandlers() {
-        // Avoid recursion if there are bugs in the logging code.
-        let inLogger = false;
-
-        const logUnhandledException = (err) => {
-            if (!inLogger) {
-                inLogger = true;
-                if (!err.stack) {
-                    Error.captureStackTrace(err);
-                }
-                this.log('fatal/service-runner/unhandled', err);
-                inLogger = false;
-            }
-        };
-
-        // Catch unhandled rejections & log them. This relies on bluebird.
-        process.on('unhandledRejection', logUnhandledException);
-
-        // Similarly, log uncaught exceptions. Also, exit.
-        process.on('uncaughtException', (err) => {
-            logUnhandledException(err);
-            // Don't exit right away, allow logger to process message
-            setTimeout(() => {
-                process.exit(1);
-            }, 100);
-        });
     }
 
     _levelToMatcher(level) {
@@ -376,6 +345,10 @@ class Logger {
     }
 
     close() {
+        if (this._logger && this._errorHandler && this._logger.removeListener) {
+                this._logger.removeListener('error', this._errorHandler);
+        }
+
         this._logger.streams.filter((stream) => stream.type === 'file').forEach((stream) => {
             stream.stream.end();
         });
--- a/lib/master.js
+++ b/lib/master.js
@@ -35,7 +35,26 @@ class Master extends BaseService {
         this._inRollingRestart = false;
         this._firstWorkerStarted = false;
         this._firstWorkerStartupAttempts = 0;
-        this.workerStatusMap = {};
+        this._workerStatusMap = {};
+        this._shutdownMasterHandler = () => {
+            this.stop()
+            .then(() => {
+                this._logger.log('info/service-runner/master', 'Exiting master');
+                return this._exitProcess(0);
+            });
+        };
+
+        this._rollingRestartHandler = () => {
+            this._firstWorkerStarted = false;
+            this._firstWorkerStartupAttempts = 0;
+            return this._updateConfig()
+            .then(() => {
+                // Recreate loggers
+                this._logger.close();
+                this._logger = new Logger(this.config.logging);
+            })
+            .then(this._rollingRestart.bind(this));
+        };
     }
 
     _getConfigUpdateAction(conf) {
@@ -47,6 +66,12 @@ class Master extends BaseService {
             // No workers needed, run worker code directly.
             return Worker.prototype.stop.call(this);
         }
+
+        // Remove signal handlers
+        process.removeListener('SIGINT', this._shutdownMasterHandler);
+        process.removeListener('SIGTERM', this._shutdownMasterHandler);
+        process.removeListener('SIGHUP', this._rollingRestartHandler);
+
         super.stop();
         if (this.interval) {
             clearInterval(this.interval);
@@ -70,29 +95,11 @@ class Master extends BaseService {
         this._logger.log('info/service-runner',
             `master(${process.pid}) initializing ${this.config.num_workers} workers`);
 
-        const shutdownMaster = () => {
-            this.stop()
-            .then(() => {
-                this._logger.log('info/service-runner/master', 'Exiting master');
-                return this._exitProcess(0);
-            });
-        };
-
-        process.on('SIGINT', shutdownMaster);
-        process.on('SIGTERM', shutdownMaster);
+        process.on('SIGINT', this._shutdownMasterHandler);
+        process.on('SIGTERM', this._shutdownMasterHandler);
 
         // Set up rolling restarts
-        process.on('SIGHUP', () => {
-            this._firstWorkerStarted = false;
-            this._firstWorkerStartupAttempts = 0;
-            return this._updateConfig()
-            .then(() => {
-                // Recreate loggers
-                this._logger.close();
-                this._logger = new Logger(this.config.logging);
-            })
-            .then(this._rollingRestart.bind(this));
-        });
+        process.on('SIGHUP', this._rollingRestartHandler);
 
         let ratelimiterSetup = P.resolve();
         if (this.config.ratelimiter) {
@@ -124,29 +131,31 @@ class Master extends BaseService {
 
     _stopWorker(workerPid) {
         const worker = cluster.workers[workerPid];
+
         if (!worker || worker.state === 'disconnected') {
             if (worker) {
-                delete this.workerStatusMap[worker.process.pid];
+                delete this._workerStatusMap[worker.process.pid];
             }
             return;
         }
-        this.workerStatusMap[worker.process.pid] = {
+        this._workerStatusMap[worker.process.pid] = {
             time: null,
             killed: true
         };
+
         const res = new P((resolve) => {
             const timeout = setTimeout(() => {
                 // worker.kill doesn't send a signal immediately, it waits until
                 // worker closes all connections with master. If after a minute
                 // it didn't happen, don't expect it happen ever.
                 worker.process.kill('SIGKILL');
-                delete this.workerStatusMap[worker.process.pid];
+                delete this._workerStatusMap[worker.process.pid];
                 resolve();
             }, 60000);
             worker.once('disconnect', () => {
                 clearTimeout(timeout);
                 worker.process.kill('SIGKILL');
-                delete this.workerStatusMap[worker.process.pid];
+                delete this._workerStatusMap[worker.process.pid];
                 resolve();
             });
         });
@@ -155,9 +164,9 @@ class Master extends BaseService {
     }
 
     _onStatusReceived(worker, status) {
-        const val = this.workerStatusMap[worker.process.pid] || {};
+        const val = this._workerStatusMap[worker.process.pid] || {};
         val.status = status;
-        this.workerStatusMap[worker.process.pid] = val;
+        this._workerStatusMap[worker.process.pid] = val;
     }
 
     _startWorker(workerId) {
@@ -172,10 +181,16 @@ class Master extends BaseService {
                 type: 'config',
                 body: yaml.dump(config)
             });
-            const startupWorkerExit = (code) => {
+
+            let workerMessageHandler = null;
+
+            const startupWorkerExitHandler = (code) => {
                 if (this._shuttingDown || this._inRollingRestart) {
                     return;
                 }
+
+                worker.removeListener('exit', startupWorkerExitHandler);
+                worker.removeListener('message', workerMessageHandler);
 
                 if (!this._firstWorkerStarted &&
                         this._firstWorkerStartupAttempts++ >= STARTUP_ATTEMPTS_LIMIT) {
@@ -186,6 +201,7 @@ class Master extends BaseService {
                         worker_pid: worker.process.pid,
                         exit_code: code
                     });
+
                     return this._exitProcess(1);
                 }
 
@@ -207,8 +223,9 @@ class Master extends BaseService {
                     resolve(this._startWorker(workerId));
                 });
             };
+            worker.on('exit', startupWorkerExitHandler);
 
-            const workerExit = (worker) => {
+            const workerExitHandler = (worker) => {
                 if (this._shuttingDown || this._inRollingRestart) {
                     return;
                 }
@@ -218,24 +235,26 @@ class Master extends BaseService {
                     worker_pid: worker.process.pid,
                     exit_code: worker.process.exitCode
                 };
-                if (this.workerStatusMap[worker.process.pid] &&
-                    this.workerStatusMap[worker.process.pid].status) {
-                    info.status = this.workerStatusMap[worker.process.pid].status;
+                if (this._workerStatusMap[worker.process.pid] &&
+                    this._workerStatusMap[worker.process.pid].status) {
+                    info.status = this._workerStatusMap[worker.process.pid].status;
                 }
                 this._logger.log('error/service-runner/master', info);
-                delete this.workerStatusMap[worker.process.pid];
+                delete this._workerStatusMap[worker.process.pid];
+                worker.removeListener('exit', workerExitHandler);
+                worker.removeListener('message', workerMessageHandler);
+
                 P.delay(Math.random() * 2000).then(() => {
                     resolve(this._startWorker(worker.worker_id));
                 });
             };
 
-            worker.on('exit', startupWorkerExit);
-            worker.on('message', (msg) => {
+            workerMessageHandler = (msg) => {
                 switch (msg.type) {
                     case 'startup_finished':
-                        worker.removeListener('exit', startupWorkerExit);
+                        worker.removeListener('exit', startupWorkerExitHandler);
                         worker.on('exit', () => {
-                            workerExit(worker);
+                            workerExitHandler(worker);
                         });
                         this._firstWorkerStarted = true;
                         resolve(msg.serviceReturns);
@@ -253,7 +272,9 @@ class Master extends BaseService {
                         this._logger.log('error/service-runner/master',
                             `unknown message type received from worker ${msg.type}`);
                 }
-            });
+            };
+
+            worker.on('message', workerMessageHandler);
         });
     }
     // Fork a single worker, wait for it to start executing and set everything up,
@@ -278,7 +299,7 @@ class Master extends BaseService {
                 const now = new Date();
                 Object.keys(cluster.workers).forEach((workerPid) => {
                     const worker = cluster.workers[workerPid];
-                    const lastBeat = this.workerStatusMap[worker.process.pid];
+                    const lastBeat = this._workerStatusMap[worker.process.pid];
                     if (!lastBeat || (!lastBeat.killed && now - lastBeat.time >
                             this.config.worker_heartbeat_timeout)) {
                         const info = {
@@ -303,13 +324,13 @@ class Master extends BaseService {
      * @private
      */
     _saveBeat(worker) {
-        const currentVal = this.workerStatusMap[worker.process.pid];
+        const currentVal = this._workerStatusMap[worker.process.pid];
         if (currentVal && currentVal.killed) {
             return;
         }
-        this.workerStatusMap[worker.process.pid] = currentVal || {};
-        this.workerStatusMap[worker.process.pid].time = new Date();
-        this.workerStatusMap[worker.process.pid].killed = false;
+        this._workerStatusMap[worker.process.pid] = currentVal || {};
+        this._workerStatusMap[worker.process.pid].time = new Date();
+        this._workerStatusMap[worker.process.pid].killed = false;
     }
 }
 
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -20,12 +20,21 @@ const RateLimiterNoCluster = require('./ratelimiter').nocluster;
  * @constructor
  */
 class Worker extends BaseService {
+    constructor() {
+        super();
+
+        this._shutdownHandler = null;
+        this._dumpHeapHandler = null;
+        this._messageHandler = null;
+        this._serviceStatusHandler = null;
+    }
+
     _getConfigUpdateAction() {
         return new P((resolve, reject) => {
             const timeout = setTimeout(() => {
                 reject(new Error(`Timeout waiting for config in worker ${process.pid}`));
             }, 3000);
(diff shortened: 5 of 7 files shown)


lib/heapwatch.js
```diff
@@ -107,6 +110,7 @@ class HeapWatch {
     }
 
     close() {
+        this.gcStats.removeListener('stats', this.reportStatsHandler);
```
Pchelolo: The 'require' of the gc-stats package was surrounded with try-catch because it's an optional binary dependency, thus requiring it might error out, thus the gcStats will be null, thus this will fail during shutdown.

lib/logger.js
```diff
@@ -112,8 +111,37 @@ class Logger {
             this._traceLogger = this._logger.child({
                 level: bunyan.TRACE
             });
-            // Set up handlers for uncaught extensions
-            this._setupRootHandlers();
+
+            // Avoid recursion if there are bugs in the logging code.
+            this.inLogger = false;
+
+            this.logUnhandledException = (err) => {
```
Pchelolo: This should be moved up to the `base_service` as well. The idea is that we are using the logger here, but doing something with unhandled exceptions is not the task of a logger itself.

Also, the `inLogger` recusrion protection was lost for `unhandledRejection` handler.

lib/logger.js
```diff
@@ -112,8 +111,37 @@ class Logger {
...
+                if (!this.inLogger) {
+                    this.inLogger = true;
+                    this.log('fatal/service-runner/unhandled', err);
+                    this.inLogger = false;
+                }
+            };
+
+            this.errorHandler = (err, failedStream) => {
```
Pchelolo: On the countrary to the previous comment, this is internal dealings of the logger itself, so it's correct that it was left alone here. But why was the logic of adding/removing it was moved up? it should stay here as it was. The logger has a _close method which would be the perfect spot for removing any internal handlers the logger sets up for itself. 

Plus, it's a private matter, let's prefix the field name with `_`

lib/worker.js
```diff
@@ -20,12 +20,21 @@ const RateLimiterNoCluster = require('./ratelimiter').nocluster;
  * @constructor
  */
 class Worker extends BaseService {
+    constructor() {
+        super();
+
+        this.shutdownHandler = null;
+        this.dumpHeapHandler = null;
+        this.messageHandler = null;
+        this.serviceStatusHandler = null;
```
Pchelolo: Let's prefix all of these with `_`

lib/worker.js
```diff
@@ -20,12 +20,21 @@ const RateLimiterNoCluster = require('./ratelimiter').nocluster;
...
+    }
+
     _getConfigUpdateAction() {
         return new P((resolve, reject) => {
             const timeout = setTimeout(() => {
                 reject(new Error(`Timeout waiting for config in worker ${process.pid}`));
             }, 3000);
-            process.on('message', (message) => {
+            process.on('message', this.messageHandler = function (message) {
```
Pchelolo: Why chaging it to `function`? It will create a new scope for `this` and screw line 40

lib/worker.js
```diff
@@ -41,7 +50,7 @@ class Worker extends BaseService {
 
             if (cluster.isWorker) {
                 // If got a status update in a worker - forward it to the master
-                process.on('service_status', (message) => {
+                process.on('service_status', this.serviceStatusHandler = function (message) {
```
Pchelolo: Idem

lib/worker.js
```diff
@@ -82,20 +106,22 @@ class Worker extends BaseService {
...
-            this._exitProcess(0);
-        }));
+        });
 
         // Enable heap dumps in /tmp on kill -USR2.
         // See https://github.com/bnoordhuis/node-heapdump/
         // For node 0.6/0.8: npm install heapdump@0.1.0
         // For 0.10: npm install heapdump
-        process.on('SIGUSR2', () => {
+        process.on('SIGUSR2', this.dumpHeapHandler = function () {
```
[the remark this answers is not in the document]
hknustwmf: Need a reference to the function to remove. Anonymous is the issue here

lib/worker.js
```diff
@@ -82,20 +106,22 @@ class Worker extends BaseService {
 
     _start() {
         // Worker.
-        process.on('SIGTERM', () => this.stop()
-        .then(() => {
-            this._logger.log('info/service-runner/worker', {
-                message: 'worker shutting down',
-                worker_pid: process.pid
+        process.on('SIGTERM', this.shutdownHandler = function () {
```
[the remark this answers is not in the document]
hknustwmf: idem. can carry the scope over if necessary

lib/base_service.js
```diff
@@ -10,6 +10,10 @@ const Logger = require('./logger');
 const docker = require('./docker');
 
 const NUM_CPU_REGEX = /^(?:(?:ncpu[\s)*+-/])|[\s()*+-/.\d])+(?:ncpu)?$/;
+
+global.uncaughtExceptionHandlerAdded = false;
+global.unhandledRejectionHandlerAdded = false;
```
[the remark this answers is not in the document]
hknustwmf: The events are tied to the process. Single process, single event. Lets discuss in the 1-1

lib/base_service.js
```diff
@@ -32,12 +36,29 @@ class BaseService {
         this._metrics = null;
         this._ratelimiter = null;
         this.serviceReturns = null;
+
+        this.uncaughtExceptionHandler = (err) => {
```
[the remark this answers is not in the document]
hknustwmf: No problem

lib/heapwatch.js
hknustwmf: Makes sense. I'll add a check for it.

lib/logger.js
hknustwmf: I'll take another look. The logUnhandledException made sense to me in the logger module

lib/logger.js
hknustwmf: Ok

lib/worker.js
hknustwmf: Ok

lib/worker.js
hknustwmf: idem

lib/worker.js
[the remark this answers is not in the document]
Pchelolo:
```
const EventEmitter = require('events').EventEmitter;
class A extends EventEmitter {
    setup() {
        this.addListener('event', this._log = () => {
            console.log('LOOOOGGGGG!!!');
        });
    }
    teardown() {
        this.removeListener('event', this._log);
    }
}
console.log('create');
const a = new A();
console.log('emit');
a.emit('event'); // Does not logs
console.log('setup');
a.setup();
console.log('emit');
a.emit('event'); // logs
console.log('teardown');
a.teardown();
console.log('emit');
a.emit('event'); // Does not emit
```

```
create
emit
setup
emit
LOOOOGGGGG!!!
teardown
emit
```

Works perfectly with arrow functions... Retaining a reference to it is enough, AFAIK it doesn't matter which syntax is used.

lib/base_service.js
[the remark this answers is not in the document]
hknustwmf: Done

lib/logger.js
hknustwmf: Moved to base_service.js

lib/worker.js
hknustwmf: Changed to arrow function

lib/worker.js
hknustwmf: Changed to arrow function

lib/base_service.js
```diff
@@ -31,13 +32,48 @@ class BaseService {
...
+            this._uncaughtExceptionHandler = (err) => {
+                this._logUnhandledException(err);
+                // Don't exit right away, allow logger to process message
+                setTimeout(() => {
+                    process.exit(1);
+                }, 100);
+            };
+        }
+        // Avoid recursion if there are bugs in the logging code.
+        this.inLogger = false;
```
Pchelolo: Let's prepend with `_`, it's a private field.

lib/base_service.js
```diff
@@ -31,13 +32,48 @@ class BaseService {
         this._logger = null;
         this._metrics = null;
         this._ratelimiter = null;
-        this.serviceReturns = null;
+        this._serviceReturns = null;
+        this._uncaughtExceptionHandlerAdded = false;
+        this._unhandledRejectionHandlerAdded = false;
```
Pchelolo: We don't really need these booleans, we can just initialize the handlers to underfined.

lib/base_service.js
```diff
@@ -31,13 +32,48 @@ class BaseService {
...
         }
+        // Remove handlers
+        if (this._logger && this._logger.removeListener) {
+            if (this._logger.removeListener) {
+                this._logger.removeListener('error', this._logger.errorHandler);
+            }
+
+            process.removeListener('unhandledRejection', this._logUnhandledException);
+            process.removeListener('uncaughtException', this._uncaughtExceptionHandler);
+        }
```
Pchelolo: This doesn't reset the `unhandled*Added booleans, so whe we restart it back the new handlers will not be re-added. 

Since we don't need the booleans, we need to check if the handlers are not undefined, remove them and check whether the candlers are not undefined.

lib/base_service.js
```diff
@@ -31,13 +32,48 @@ class BaseService {
...
+        // Avoid recursion if there are bugs in the logging code.
+        this.inLogger = false;
+
+        if (!this._unhandledRejectionHandlerAdded) {
+            this._unhandledRejectionHandlerAdded = true;
+            this._logUnhandledException = (err) => {
+                if (!this.inLogger) {
+                    this.inLogger = true;
+                    this.log('fatal/service-runner/unhandled', err);
```
Pchelolo: this.log???

lib/base_service.js
```diff
@@ -31,13 +32,48 @@ class BaseService {
...
 
     stop() {
         if (this._ratelimiter) {
             this._ratelimiter.stop();
         }
+        // Remove handlers
+        if (this._logger && this._logger.removeListener) {
+            if (this._logger.removeListener) {
+                this._logger.removeListener('error', this._logger.errorHandler);
+            }
```
Pchelolo: As I have said before, this is an internal affair of the logger. The logger has a close method where this belongs. The logger.close must be called last, so it should be called from overriden stop methods of master and worker.

lib/heapwatch.js
```diff
@@ -21,28 +21,31 @@ class HeapWatch {
...
+                // Report GC timings to statsd (in nanoseconds).
+                const type = this.gcTypeName(stats.gctype);
+                if (type !== 'unknown') {
+                    this.cumulativeGCTimes[this.gcTypeName(stats.gctype)] += stats.pause;
+                }
+        };
+        this.gcStats = null;
+    }
+
+    gcTypeName(typeID) {
```
Pchelolo: This is a private method. Prepend with `_`.

lib/logger.js
Pchelolo: This is still not done.

lib/worker.js
```diff
@@ -82,20 +106,22 @@ class Worker extends BaseService {
 
     _start() {
         // Worker.
-        process.on('SIGTERM', () => this.stop()
-        .then(() => {
-            this._logger.log('info/service-runner/worker', {
-                message: 'worker shutting down',
-                worker_pid: process.pid
+        process.on('SIGTERM', this._shutdownHandler = () => {
```
Pchelolo: Why introducing a block `{}`?

Pchelolo:
Better, but not there yet. Still didn't go through the master code completely.

lib/base_service.js
```diff
@@ -91,6 +127,22 @@ class BaseService {
...
+                process.on('unhandledRejection', this._logUnhandledException);
+            }
+
+            // Similarly, log uncaught exceptions. Also, exit.
+            if (!this._uncaughtExceptionHandlerAdded) {
+                this._uncaughtExceptionHandlerAdded = true;
+                process.on('uncaughtException', this._uncaughtExceptionHandler);
+            }
+
+            if (this._logger.errorHandler && this._logger.on) {
```
Pchelolo: As I've said in the previous review, this does not belong here, its internal affears of the logger.

lib/master.js
```diff
@@ -36,6 +36,29 @@ class Master extends BaseService {
...
+                // Recreate loggers
+                this._logger.close();
+                this._logger = new Logger(this.config.logging);
+            })
+            .then(this._rollingRestart.bind(this));
+        };
+
+        this.startupWorkerExitHandlerMap = {};
+        this.workerExitHandlerMap = {};
+        this.workerMessageHandlerMap = {};
```
Pchelolo: All of these needs to be prepended with `_`

lib/master.js
```diff
@@ -134,6 +150,10 @@ class Master extends BaseService {
             time: null,
             killed: true
         };
+
+        worker.removeListener('message', this.workerMessageHandlerMap[worker.process.pid]);
+        worker.removeListener('exit', this.workerExitHandlerMap[worker.process.pid]);
+
```
Pchelolo: Haven't we just removed the same handler on the line 140?

lib/master.js
Pchelolo: If it's a map, it needs to be a Map

lib/master.js
Pchelolo: When you are removing these handlers, you are not cleanin them out of the map. memory leak.

lib/master.js
```diff
@@ -134,6 +150,10 @@ class Master extends BaseService {
             time: null,
             killed: true
         };
+
+        worker.removeListener('message', this.workerMessageHandlerMap[worker.process.pid]);
+        worker.removeListener('exit', this.workerExitHandlerMap[worker.process.pid]);
```
Pchelolo: This will break the restart logic. I still have to think a bit what to do here, but you have other comments to work on right now.

Pchelolo: Ok. Here's what we need to do. 

We need to remove the `workerExit` or `workerStartupExit` listeners from within themselves. The worker is guaranteed to exit and the 'exit' message is guaranteed to be dispatched, so it's safe. These listers have to be removed as a .finally call within themselves. Given that we can pass the references to the listener functions as a part of the closure scope, we don't need these maps. 

As for the 'message' listener - the same thing - we need to remove them within the exit listeners and we can actually pass the reference in a closure scope, so the map for that is not needed either.

Pchelolo:
I've decided it's enough to not have any tests! I've created #211 to make some integration tests and I've added a test for this. Please rebase once you are done with my comments.

lib/base_service.js
hknustwmf: Getting rid of the booleans

lib/worker.js
hknustwmf: I don't follow. We can discuss tomorrow.

Pchelolo: I mean, why did you wrap the handler in `{}`? There's no need for that and simple () => this.stop().... looks nicer.

lib/base_service.js
```diff
@@ -31,13 +32,41 @@ class BaseService {
...
+                // Don't exit right away, allow logger to process message
+                setTimeout(() => {
+                    process.exit(1);
+                }, 100);
+            };
+        }
+        // Avoid recursion if there are bugs in the logging code.
+        this._inLogger = false;
+
+        if (!this._unhandledRejectionHandler) {
```
Pchelolo: Since it's a constructor this if will always evaluate to 'true'

lib/base_service.js
```diff
@@ -31,13 +32,41 @@ class BaseService {
         this._logger = null;
         this._metrics = null;
         this._ratelimiter = null;
-        this.serviceReturns = null;
+        this._serviceReturns = null;
+        this._logUnhandledExceptionHandler = null;
+        this._unhandledRejectionHandler = null;
+
+        if (!this._logUnhandledExceptionHandler) {
```
Pchelolo: Since it's a constructor this if will always evaluate to true

lib/base_service.js
```diff
@@ -91,6 +120,18 @@ class BaseService {
             this._logger = new Logger(config.logging);
             this._setUpDNSCaching();
 
+            // Catch unhandled rejections & log them. This relies on bluebird.
+            if (!this._unhandledRejectionHandlerAdded) {
+                this._unhandledRejectionHandlerAdded = true;
+                process.on('unhandledRejection', this._logUnhandledException);
```
Pchelolo: There's still this boolean we discussed we don't need.

lib/logger.js
```diff
@@ -112,38 +111,38 @@ class Logger {
...
         } else {
             this._sampled_levels = confOrLogger._sampled_levels;
             this._logger = confOrLogger._logger;
             this._levelMatcher = confOrLogger._levelMatcher;
             this._componentLoggers = confOrLogger._componentLoggers;
             this._traceLogger = confOrLogger._traceLogger;
         }
-        this.args = args;
-    }
+        if (this._logger._errorHandler && this._logger.on) {
```
Pchelolo: Both of these conditions will evaluate to 'true' always.

lib/logger.js
```diff
@@ -112,38 +111,38 @@ class Logger {
...
+            this._logger.on('error', this._logger._errorHandler );
```
Pchelolo: We can add it on L115 too, just also store. like

`this._logger.on('error', this._errorHandler = (...) => {...}`

lib/logger.js
```diff
@@ -374,6 +347,10 @@ class Logger {
     }
 
     close() {
+        if (this._logger && this._logger.removeListener) {
```
Pchelolo: Both conditions will always evaluate to true.

lib/logger.js
```diff
@@ -374,6 +347,10 @@ class Logger {
...
+                this._logger.removeListener('error', this._logger._errorHandler);
```
Pchelolo: It's not `this._logger._errorHandler`, it's `this._errorHandler`

lib/worker.js
```diff
@@ -69,8 +78,23 @@ class Worker extends BaseService {
...
+        }
+        if (this._dumpHeapHandler) {
+            process.removeListener('SIGUSR2', this._dumpHeapHandler);
+        }
+        if (this._serviceStatusHandler) {
+            process.removeListener('service_status', this._serviceStatusHandler);
+        }
+        if (this._messageHandler) {
+            process.removeListener('message', this._messageHandler);
+        }
```
Pchelolo: We need to de-initialize the fields too and set them to undefined.

lib/base_service.js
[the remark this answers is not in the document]
hknustwmf: Removed

lib/base_service.js
hknustwmf: Done

lib/base_service.js
hknustwmf: Done

lib/base_service.js
hknustwmf: Done

lib/heapwatch.js
hknustwmf: Done

lib/base_service.js
hknustwmf: Moved to the logger close method

lib/worker.js
hknustwmf: Done

lib/base_service.js
hknustwmf: Moved to logger.js

lib/master.js
hknustwmf: Prepended with _ and renamed Dict

lib/master.js
hknustwmf: Fixed

lib/logger.js
hknustwmf: Done

lib/base_service.js
hknustwmf: Fixed

lib/base_service.js
hknustwmf: Removed

lib/base_service.js
hknustwmf: Removed

lib/logger.js
```diff
@@ -112,38 +111,38 @@ class Logger {
...
+                if (failedStream.type === 'file') {
+                    // However, if it's a `file` stream, it's likely that
+                    // we're in some catastrophic state. Remove the stream,
+                    // log the error and hope other streams would deliver it.
+                    // Attempting to continue logging through a failed `file`
+                    // stream might end up in a memory leak.
+                    this._logger.streams = this._logger.streams.filter((s) => s !== failedStream);
+                    failedStream.stream.destroy();
+                    // Hope that we have other streams to report the problem
+                    this._logger.log('fatal/service-runner/logger', {
```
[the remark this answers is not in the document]
hknustwmf: Changed

lib/logger.js
hknustwmf: I moved it up

lib/logger.js
hknustwmf: Removed

lib/logger.js
hknustwmf: Good point

lib/worker.js
hknustwmf: Done

lib/base_service.js
```diff
@@ -31,13 +32,39 @@ class BaseService {
...
+            }
+        };
     }
 
     stop() {
         if (this._ratelimiter) {
             this._ratelimiter.stop();
         }
+        // Remove handlers
+        if (this._unhandledRejectionHandler) {
```
Pchelolo: This will always evaluate to 'true'

lib/base_service.js
```diff
@@ -31,13 +32,39 @@ class BaseService {
...
+            process.removeListener('unhandledRejection', this._unhandledRejectionHandler);
+        }
+        if (this._unhandledExceptionHandler) {
```
Pchelolo: This will always evaluate to 'true'

lib/base_service.js
```diff
@@ -91,6 +118,16 @@ class BaseService {
             this._logger = new Logger(config.logging);
             this._setUpDNSCaching();
 
+            // Catch unhandled rejections & log them. This relies on bluebird.
+            if (!this._unhandledRejectionHandler) {
```
Pchelolo: This will always evaluate to 'false'

lib/base_service.js
```diff
@@ -91,6 +118,16 @@ class BaseService {
...
+                process.on('unhandledRejection', this._unhandledRejectionHandler);
+            }
+
+            // Similarly, log uncaught exceptions. Also, exit.
+            if (!this._uncaughtExceptionHandler) {
```
Pchelolo: This will always evaluate to 'false'

lib/logger.js
```diff
@@ -376,6 +373,10 @@ class Logger {
     }
 
     close() {
+        if (this._logger && this._errorHandler && this._logger.removeListener) {
```
Pchelolo: Hm. At first I was thinking this is too many conditions, but actually, it's not. The error handler will only be set up for the root logger and not the child loggers. but given that all the logger children share an internal Bunyan logger, calling a close method of a child will destroy the whole hierarchy. No need to take action now, it's a separate issue I've just noticed and we will deal with it separately.

lib/master.js
```diff
@@ -35,7 +35,30 @@ class Master extends BaseService {
...
+                // Recreate loggers
+                this._logger.close();
+                this._logger = new Logger(this.config.logging);
+            })
+            .then(this._rollingRestart.bind(this));
+        };
+
+        this._startupWorkerExitHandler = null;
+        this._workerExitHandler = null;
+        this._workerMessageHandler = null;
```
Pchelolo: Nonononono... We don't need them to be properties of the master object. Now with many workers we will just override this variable as we start them so it will only remove handlers from the last started worker. These need to be local variable in `_startWorker` function and be confined to it's closure.

lib/master.js
```diff
@@ -172,7 +188,7 @@ class Master extends BaseService {
                 type: 'config',
                 body: yaml.dump(config)
             });
-            const startupWorkerExit = (code) => {
+            this._startupWorkerExitHandler = (code) => {
```
Pchelolo: It was correct how it was.

lib/master.js
```diff
@@ -207,8 +223,9 @@ class Master extends BaseService {
                     resolve(this._startWorker(workerId));
                 });
             };
+            worker.on('exit', this._startupWorkerExitHandler);
 
-            const workerExit = (worker) => {
+            this._workerExitHandler = (worker) => {
```
Pchelolo: It was correct how it was.

lib/master.js
```diff
@@ -218,24 +235,25 @@ class Master extends BaseService {
...
-                delete this.workerStatusMap[worker.process.pid];
+                delete this._workerStatusMap[worker.process.pid];
                 P.delay(Math.random() * 2000).then(() => {
                     resolve(this._startWorker(worker.worker_id));
                 });
             };
 
-            worker.on('exit', startupWorkerExit);
-            worker.on('message', (msg) => {
+            this._workerMessageHandler = (msg) => {
```
Pchelolo: It has to be a local variable.

lib/master.js
```diff
@@ -218,24 +235,25 @@ class Master extends BaseService {
...
                 switch (msg.type) {
                     case 'startup_finished':
-                        worker.removeListener('exit', startupWorkerExit);
+                        if (this._startupWorkerExitHandler) {
```
Pchelolo: No need for this 'if' afaik.

lib/master.js
```diff
@@ -218,24 +235,25 @@ class Master extends BaseService {
...
+                            worker.removeListener('exit', this._startupWorkerExitHandler);
+                        }
                         worker.on('exit', () => {
-                            workerExit(worker);
+                            this._workerExitHandler(worker);
```
Pchelolo: it was good how it was.

Pchelolo:
Almost. Left some more comments. Plus, you have not unskipped the test.

lib/base_service.js
hknustwmf: Removed. This is the result of  code change, I had original set the property to null.

lib/base_service.js
hknustwmf: Same issue removed.

lib/base_service.js
hknustwmf: Same issue. removed.

lib/logger.js
hknustwmf: Ok

lib/master.js
hknustwmf: Reverted

lib/master.js
hknustwmf: Removed

lib/master.js
hknustwmf: workerExit is removed in _stopWorker. So the local var would be out scope.

Pchelolo: And as we have discussed it's incorrect to remove it there.

lib/master.js
hknustwmf: That's why I had a map there in the first place to for handlers.

Pchelolo: I understood that. What I'm saying is that you don't need any of those to be properties of a master  object. These all need to be local variables.

lib/master.js
hknustwmf: Same issue as with workerExit

lib/base_service.js
hknustwmf: Removed

hknustwmf:
> Almost. Left some more comments. Plus, you have not unskipped the test.

I don't remember you telling me too. Will do.

lib/master.js
Pchelolo: And as I have explained, we don't need to do it there. The workerExit must be removed from within a workerExit

lib/master.js
Pchelolo: And same issue as with workerExit - we don't need to remove it from stopWorker, we need to remove it from workerExit.

I thought you said you understood my explanations on our meetings yesterday and it was clear how it all fits together. If it was not clear, why didn't you ask more followup questions?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "service-runner",
-  "version": "2.6.18",
+  "version": "2.6.19",
```
Pchelolo: Sorry, already released .19 cause we were blocking parsing. It's a big one, let's give it a minor version bump?

lib/master.js
```diff
@@ -207,8 +216,11 @@ class Master extends BaseService {
                     resolve(this._startWorker(workerId));
                 });
             };
+            worker.on('exit', startupWorkerExitHandler);
 
-            const workerExit = (worker) => {
+            var workerMessageHandler = null;
```
Pchelolo: let?

package.json
hknustwmf: Changed it 2.7.0

lib/master.js
hknustwmf: Bad old habits. Block scope is enough

Pchelolo:
Simply updating the version creates a merge conflict, as the version was updated in the master. You need to rebase now.

lib/base_service.js
```diff
@@ -31,13 +32,46 @@ class BaseService {
...
+                process.exit(1);
+            }, 100);
+        };
+
+        this._unhandledRejectionHandler = (err) => {
+            if (!this._inLogger) {
+                this._inLogger = true;
+                this._logger.log('fatal/service-runner/unhandled', err);
+                this._inLogger = false;
+            }
```
Pchelolo: I believe that before this change both handlers were essentially calling the same `logUnhandledException` code, git the unhandled exception handler was exiting the process as well. Can you please elaborate on why you have changed that?

Also, if you do that you only need to declare 2 functions, not 3.

lib/master.js
```diff
@@ -172,7 +181,7 @@ class Master extends BaseService {
                 type: 'config',
                 body: yaml.dump(config)
             });
-            const startupWorkerExit = (code) => {
+            const startupWorkerExitHandler = (code) => {
```
Pchelolo: We need to remove the startupWorkerExitHandler from within the startupWorkerExitHandler as well

Pchelolo: Same as the workerMessageHandler.

lib/master.js
```diff
@@ -218,24 +230,26 @@ class Master extends BaseService {
...
-            worker.on('exit', startupWorkerExit);
-            worker.on('message', (msg) => {
+            workerMessageHandler = (msg) => {
                 switch (msg.type) {
                     case 'startup_finished':
-                        worker.removeListener('exit', startupWorkerExit);
+                        worker.removeListener('exit', startupWorkerExitHandler);
                         worker.on('exit', () => {
-                            workerExit(worker);
+                            workerExitHandler(worker);
```
Pchelolo: No action needed. TODO: mental note for myself - we can use the worker in the exit handler from the closure and pass in the exit code to log it as well.

Pchelolo:
Ok, the last little set of comments, I promise :) Missed a couple of things in previous review

lib/base_service.js
hknustwmf: I changed it to 2. That was by mistake.

lib/master.js
hknustwmf: Added both

lib/base_service.js
```diff
@@ -32,13 +33,38 @@ class BaseService {
...
         this._metrics = null;
         this._ratelimiter = null;
-        this.serviceReturns = null;
+        this._serviceReturns = null;
+
+        // Avoid recursion if there are bugs in the logging code.
+        this._inLogger = false;
+
+        this._uncaughtExceptionHandler = (err) => {
+            this.logUnhandledException(err);
```
Pchelolo: Em, man... You've just removed this metod!

Pchelolo:
One more...

lib/base_service.js
```diff
@@ -32,13 +33,38 @@ class BaseService {
         this._logger = null;
         this._metrics = null;
         this._ratelimiter = null;
-        this.serviceReturns = null;
+        this._serviceReturns = null;
+
+        // Avoid recursion if there are bugs in the logging code.
+        this._inLogger = false;
+
+        this._uncaughtExceptionHandler = (err) => {
```
d00rman: Why are we declaring this method inside the constructor? Let's be consistent and declare it just like all other methods.

lib/base_service.js
```diff
@@ -32,13 +33,38 @@ class BaseService {
...
+            this.logUnhandledException(err);
+            // Don't exit right away, allow logger to process message
+            setTimeout(() => {
+                process.exit(1);
+            }, 100);
+        };
+
+        this._unhandledRejectionHandler = (err) => {
```
d00rman: Idem.

lib/base_service.js
```diff
@@ -92,6 +118,12 @@ class BaseService {
             this._logger = new Logger(config.logging);
             this._setUpDNSCaching();
 
+            // Catch unhandled rejections & log them. This relies on bluebird.
+            process.on('unhandledRejection', this._unhandledRejectionHandler);
```
d00rman: Wouldn't it be better to declare these right from the get-go instead of waiting for everything else to be set-up? Doing so would improve start-up logging itself too.

lib/heapwatch.js
```diff
@@ -21,28 +21,31 @@ class HeapWatch {
         this.timeoutHandle = undefined;
         this.gcReportInterval = undefined;
         this.cumulativeGCTimes = Object.assign({}, ZERO_CUMULATIVE_GC_INTERVAL);
+        this.reportStatsHandler = (stats) => {
```
d00rman: Idem as for BaseService. Let's not hide method declarations like this.

lib/heapwatch.js
```diff
@@ -21,28 +21,31 @@ class HeapWatch {
...
-                }
-            };
-            gcStats.on('stats', (stats) => {
-                // Report GC timings to statsd (in nanoseconds).
-                const type = gcTypeName(stats.gctype);
-                if (type !== 'unknown') {
-                    this.cumulativeGCTimes[gcTypeName(stats.gctype)] += stats.pause;
-                }
-            });
+            this._gcStats = require('gc-stats')();
```
d00rman: This `require` needs to be guarded, as it's not guaranteed that `gc-stats` is available, and we don't want the process to crash entirely if that is the case.

lib/logger.js
```diff
@@ -112,38 +111,36 @@ class Logger {
             this._traceLogger = this._logger.child({
                 level: bunyan.TRACE
             });
-            // Set up handlers for uncaught extensions
-            this._setupRootHandlers();
+
+            this._errorHandler = (err, failedStream) => {
```
d00rman: Idem...

lib/master.js
```diff
@@ -35,7 +35,26 @@ class Master extends BaseService {
         this._inRollingRestart = false;
         this._firstWorkerStarted = false;
         this._firstWorkerStartupAttempts = 0;
-        this.workerStatusMap = {};
+        this._workerStatusMap = {};
+        this._shutdownMasterHandler = () => {
```
d00rman: Idem

lib/master.js
```diff
@@ -35,7 +35,26 @@ class Master extends BaseService {
...
+            this.stop()
+            .then(() => {
+                this._logger.log('info/service-runner/master', 'Exiting master');
+                return this._exitProcess(0);
+            });
+        };
+
+        this._rollingRestartHandler = () => {
```
d00rman: Idem

lib/master.js
```diff
@@ -172,11 +181,17 @@ class Master extends BaseService {
                 type: 'config',
                 body: yaml.dump(config)
             });
-            const startupWorkerExit = (code) => {
+
+            let workerMessageHandler = null;
+
+            const startupWorkerExitHandler = (code) => {
```
d00rman: Idem

lib/master.js
```diff
@@ -207,8 +223,9 @@ class Master extends BaseService {
                     resolve(this._startWorker(workerId));
                 });
             };
+            worker.on('exit', startupWorkerExitHandler);
 
-            const workerExit = (worker) => {
+            const workerExitHandler = (worker) => {
```
d00rman: Idem

lib/base_service.js
Pchelolo: No. That's actually correct. If we declare the method like any other method, it will be a property of the object prototype thus shared between all instances of the master/worker. So we wouldn't be able to create several service-runners within the same process and add/remove appropriate event handlers on start/stop.

lib/base_service.js
Pchelolo: Idem

lib/base_service.js
Pchelolo: Si, Agreed.

lib/heapwatch.js
Pchelolo: Idem to that answer in the base_service.

lib/base_service.js
hknustwmf: Fixed

lib/heapwatch.js
hknustwmf: I am not sure I follow. The require is wrapped in a try/catch.

lib/logger.js
hknustwmf: Could you elaborate what you want to do here?

lib/heapwatch.js
d00rman: Oops, my bad. I don't know how I missed it.

lib/logger.js
d00rman: Ah no, nothing, this needs to stay here.

lib/base_service.js
d00rman: Could be, but the fix hasn't been uploaded.

Usually, the way we work is that we first change/address all of the comments in code, then upload the new version and then comment with the outcomes for each specific comment. That way, it's easy to follow what is happening. In this instance, you say you fixed it, but you haven't uploaded the fix itself. At the same time, there are some other comments to be addressed throughout the code, so we don't know what is really happening.

hknustwmf: I was working on something else while I was waiting for feed back. I had not marked this resolved just added comment.

d00rman: And that is fine. My point was that it's better to do a pass on all of the comments that are clear/fixed once you uploaded the fixes/changes, as that makes it less messy and easier to follow. Otherwise, both the author and reviewers can eaily tell which comments have been addressed and which haven't.

hknustwmf: Just pushed the change and mark this resolved

d00rman:
I must say that I'm getting more and more uncomfortable with the size and amount of changes in this PR. @hknustwmf please split it in at least two PRs: one that addresses logging issues and another one that addresses signal handling.

lib/base_service.js
d00rman: I saw you pushed the change, but that was beside the point of my comment. Please re-read this whole thread.

Pchelolo:
> I must say that I'm getting more and more uncomfortable with the size and amount of changes in this PR. @hknustwmf please split it in at least two PRs: one that addresses logging issues and another one that addresses signal handling.

I tend to disagree. One issue - one PR. We have done very extensive amount of work to make sure this is all correct, and, quite frankly, the changes are quite obvious and easy to follow - it's mostly variable renamings and copy-pasting some code. Thus, I don't think we need to do any more work just to make it smaller or split

PR #213, opened 2019-05-18
Don't override error stack if it is provided and rely on error code.

Pchelolo:


lib/base_service.js
```diff
@@ -362,8 +362,9 @@ class BaseService {
         try {
             return P.resolve(require(modName));
         } catch (e) {
-            if (/^\//.test(opts.mod) || (opts.baseTried && opts.modsTried) ||
-                    !e.message.startsWith(`Cannot find module '${modName}'`)) {
+            if (path.isAbsolute(opts.mod) ||
+                    (opts.baseTried && opts.modsTried) ||
+                    e.code !== 'MODULE_NOT_FOUND') {
```
d00rman: I seem to vaguely remember that we used the error message instead of the code for a reason, but can't recall it now. Why are you changing this here now?

Pchelolo: Before node 4 those codes were unstable and messages were more stable. now it's vice-versa.

d00rman: Hmmm, Node 6 docs don't seem to list that code at all (and says that all error codes start with `E`), Node 10 calls this `ERR_MISSING_MODULE` and Node 12 calls it `ERR_MODULE_NOT_FOUND` and the code itself is marked experimental there. I think we should keep this as-is for now.

PR #217, opened 2019-06-10
Do not try to set the stack trace of undefined objects; rel v2.7.1

d00rman:
Sometimes we can get `undefined` as the error object to log. When that's the case, don't attempt to add the stack trace, nor log it (since bunyan doesn't log empty messages).

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -47,9 +47,9 @@ class BaseService {
         };
 
         this._unhandledRejectionHandler = (err) => {
-            if (!this._inLogger) {
+            if (!this._inLogger && err) {
                 this._inLogger = true;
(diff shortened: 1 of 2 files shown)


Pchelolo:
Ok, there's much more I would like to do here, but just for stability we can merge it now.

Ideas for later investigation:
0. Why are we getting undefined in the first place?
1. If the `err` is not guaranteed to be `err` - what else could we do? Wrap it in `Error`?
2. The `unhandledRejection` also gets the instance of the original promise. Can we inspect it to find more info about the root cause?

I'll investingate these.

d00rman:
Published as [v2.7.1](https://github.com/wikimedia/service-runner/tree/v2.7.1)

PR #218, opened 2019-06-12
Docker: Update to NVM v0.34.0

d00rman:


d00rman:
Trivial, self-merged.

PR #220, opened 2019-07-31
Refactor shutdown sequence; rel v2.7.3

Pchelolo:
1. Update dependencies.
2. Don't attempt to send heartbeats if we're shutting down.
3. Refactor shutdown sequence. Before we were killing the worker when master detected the worker was disconnected. That's actually no good, cause master can emit the `disconnect` event before the worker does, so this doesn't leave the worker required time to gracefully kill itself after all the cleanup is done. Now we only SIGKILL the worker if it's stuck in its shutdown sequence for some time, and normally relying on the worker to correctly kill itself.

Bug: https://phabricator.wikimedia.org/T158265

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -30,6 +30,8 @@ class BaseService {
 
         this.options = this._getOptions(options);
 
+        this._running = false;
+
         this._logger = null;
         this._metrics = null;
         this._ratelimiter = null;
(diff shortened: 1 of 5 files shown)


d00rman:
Published as [v2.7.3](https://github.com/wikimedia/service-runner/tree/v2.7.3)

PR #221, opened 2019-09-05
initial attempt at implementing native prometheus metrics

shdubsh:
More information on ticket: https://phabricator.wikimedia.org/T205870

Apologies in advance if there are some boneheaded errors, JS is not my native tongue.  Feedback is greatly appreciated.

It should be backwards compatible except in configuration format, but both formats could exist simultaneously given no name collisions.  It seems to test ok with the simple_server module.  Tested against node v10.15.3.

Thanks!


config.yaml
```diff
@@ -22,14 +22,20 @@ logging:
...
-  type: statsd
-  host: localhost
-  port: 8125
+  prometheus:
+    type: prometheus
+    port: 9000
+  statsd:
+    type: statsd
+    host: localhost
+    port: 8125
```
Pchelolo: hm.... changing the config spec is not really great... This means that whenever a service updates to a new service-runner version, it needs to update it's config, while it might not yet switch to prometheus for actual metrics collection. Can we still support the old config and maybe log a deprecation warning if it's detected?

lib/prometheus.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const Prometheus = require('prom-client');
+
+function makePrometheusClient() {
+    return Prometheus;
+}
+
+module.exports = makePrometheusClient;
```
Pchelolo: What is the point of this file?

lib/master.js
```diff
@@ -267,6 +287,8 @@ class Master extends BaseService {
                     case 'ratelimiter_counters':
                         return this._ratelimiter &&
                                 this._ratelimiter.updateCounters(msg.value);
+                    case 'prom-client:getMetricsRes':
+                        return;
```
Pchelolo: I guess it's needed as it's an internal communication mode of the prometheus lib, but it's confusing. Let's add a comment.

lib/master.js
```diff
@@ -112,6 +118,20 @@ class Master extends BaseService {
...
+        HTTP.createServer((req, res) => {
+            AggregatorRegistry.clusterMetrics((err, metrics) => {
+                if (err) {
+                    this._logger.log('error/service-runner', err);
+                }
+                res.setHeader('Content-Type', AggregatorRegistry.contentType);
+                res.write(metrics);
+                res.end();
+            });
+        }).listen(this.config.metrics.prometheus.port);
```
Pchelolo: 1. Nothing ever closes the server.
2. It would probably be nice to move this into a separate file.
3. What if prometheus is not configured - this will explode.

test/utils/simple_server.js
```diff
@@ -4,7 +4,17 @@ const http = require('http');
...
+        name: 'hitcount',
+        help: 'a hit counter',
+        labelNames: ['worker_id']
+    });
+
+    const server = http.createServer((req, res) => {
+        this.hitcounter.labels(options.config.worker_id).inc();
+        options.metrics.increment(`${options.config.worker_id}.hitcount`);
+        res.end('ok\n');
+    });
```
Pchelolo: I guess since it's quite preductable with a counter, we can actually write a test. Start a server, call the endpoint several times, fetch the counter, see what we've got.

Pchelolo:
Travis is failing cause you don't close the server you created for prometheus, so the process never finishes running.

Pchelolo:
Overall looks good, left a bunch of comments mostly about backwards compatibility and how to organize things.

config.yaml
shdubsh: Great point.  It's probably best not to change the spec at all.

lib/prometheus.js
shdubsh: Something left over from pattern-matching.  Removed in latest.

lib/master.js
shdubsh: Explosions are not fun.  Please see latest.  This was removed and moved into a separate file.

test/utils/simple_server.js
shdubsh: Added a test.  I think it is sufficient, but let me know if it isn't.

config.yaml
```diff
@@ -22,14 +22,19 @@ logging:
 
 # Statsd metrics reporter
 metrics:
+  enable_prometheus: true
+  prometheus:
+    port: 9000
```
Pchelolo: Hm.. this doesn't feel great either, sorry. How about that:

metrics could be an array or an object(b-c reasons). If it's an object - treat it as a array with 1 element.

Each element describes a metrics type. So if we have both statsd and prometheus, we can have smith like 

```
metrics:
 - type: statsd
    host: localhost
    port: 8125
 - type: prometheus
    port: 9000
```

Also, we can log a deprecation warning if the config is an object.

config.yaml
```diff
@@ -22,14 +22,19 @@ logging:
 
 # Statsd metrics reporter
 metrics:
+  enable_prometheus: true
```
Pchelolo: This is not really needed, we can just check if prometheus is explicitly configured.

lib/base_service.js
```diff
@@ -390,10 +389,16 @@ class BaseService {
...
+            // And metrics clients
             if (!config.metrics.name) {
                 config.metrics.name = name;
             }
+            if (!config.metrics.prometheus) {
+                config.metrics.prometheus = {};
+            }
+            if (!config.metrics.prometheus.name) {
+                config.metrics.prometheus.name = name;
+            }
```
Pchelolo: Should we use a more confine `config.blabla = config.blabla || foobar`?

lib/master.js
```diff
@@ -91,6 +97,11 @@ class Master extends BaseService {
             .then((serviceReturns) => [serviceReturns]);
         }
 
+        // Start Prometheus Metrics Endpoint
+        if (this.config.metrics.enable_prometheus) {
+            this.metricsEndpoint = new MetricsEndpoint(this.config.metrics);
+        }
```
Pchelolo: There're 2 configurations you can run the service with - `num_workers: 0` will not use cluster mode and the whole thing will run within the master process. I think in that case your code will not work, cause the server will not be setup, cause see line 91. 

If I'm correct, we need to fix that and need to add a test.

lib/prometheus_server.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+const Prometheus = require('prom-client');
+const AggregatorRegistry = new Prometheus.AggregatorRegistry();
+const HTTP = require('http');
+
+class MetricsEndpoint {
+    constructor(config) {
+        this.server = HTTP.createServer((req, res) => {
+            AggregatorRegistry.clusterMetrics((err, metrics) => {
```
Pchelolo: This will be needed only if we're running in the cluster mode configuration? see previous comment.

test/utils/simple_config_one_worker.yaml
```diff
@@ -1,6 +1,10 @@
 num_workers: 1
```
Pchelolo: Would be good to have a test with num_workers: 0 too as that case would be quite a lot different from the cluster mode I think

test/features/tests.js
```diff
@@ -100,4 +102,33 @@ describe('service-runner tests', () => {
         })
         .finally(() => process.removeListener('warning', warningListener));
     });
+
+    it('Must increment hitcount metrics when hit', () => {
+        const server = new TestServer(`${__dirname}/../utils/simple_config_one_worker.yaml`);
+        return server.start()
+            .then(() => {
```
Pchelolo: We usually don't indent the `.then`

test/features/tests.js
```diff
@@ -100,4 +102,33 @@ describe('service-runner tests', () => {
...
+                HTTP.get('http://127.0.0.1:12345', (res) => {
+                    assert.strictEqual(res.statusCode, 200, 'Must get 200 response');
+                });
+            })
+            .delay(1000)
+            .then(() => {
+                HTTP.get('http://127.0.0.1:9000', (res) => {
```
Pchelolo: We're generally using the library `preq` for making HTTP requests. You can require it as a devDependency and use it. This would allot to simplify the test as it you wouldn't need to parse the response body manually

lib/master.js
shdubsh: This is true.  Handled in latest commit.

lib/base_service.js
shdubsh: I took a shot at making that more concise.  Let me know what you think

config.yaml
shdubsh: Took a shot at this.  Let me know what you think.

config.yaml
```diff
@@ -1,7 +1,7 @@
 # Number of worker processes to spawn.
 # Set to 0 to run everything in a single process without clustering.
 # Use 'ncpu' to run as many workers as there are CPU units
-num_workers: ncpu
+num_workers: 0
```
Pchelolo: This config is not supposed to be used for realz and is in here as an example, why are you making this change

shdubsh: Indeed.  This was left over from testing during development.  Reverted.

lib/base_service.js
```diff
@@ -390,9 +389,15 @@ class BaseService {
                 config.logging.name = name;
             }
 
-            // And the statsd client
-            if (!config.metrics.name) {
-                config.metrics.name = name;
+            // And metrics clients
+            if (!Array.isArray(config.metrics)) {
+                console.log('Deprecation warning: \'metrics\' configuration should be an array');
```
Pchelolo: console... brrrr... Unfortunately this is done way before we've set up the proper logger, so I guess it's ok..

Pchelolo:
This is good, I like it. Let's give @d00rman a chance to do a final review

Still needs to be done:
- bump a minor version.
- it would be nice to run tests for some services with this version of service-runner to make sure it's all good. RESTBase and mobile-content-service should be enough.

shdubsh:
> This is good, I like it. Let's give @d00rman a chance to do a final review
> 
> Still needs to be done:
> 
> * bump a minor version.
> * it would be nice to run tests for some services with this version of service-runner to make sure it's all good. RESTBase and mobile-content-service should be enough.

Regarding the minor version bump, is that something I should prepare in this PR or something the person releasing will do?

Pchelolo:
Just put it in the same PR. I'll publish the package once it's merged.

config.yaml
```diff
@@ -20,16 +20,20 @@ logging:
...
+    port: 9000
+  - type: statsd
+    host: localhost
+    port: 8125
 
 services:
   - name: test
     # a relative path or the name of an npm package, if different from name
-    module: ./some_server_module.js
+
```
d00rman: Remove the empy line and the comment below, no need for either of these to pollute the config.

lib/base_service.js
```diff
@@ -390,9 +389,15 @@ class BaseService {
                 config.logging.name = name;
             }
 
-            // And the statsd client
-            if (!config.metrics.name) {
-                config.metrics.name = name;
+            // And metrics clients
+            if (!Array.isArray(config.metrics)) {
```
d00rman: This whole block should be:

```js
if (!Array.isArray(config.metrics)) {
  config.metrics = config.metrics || {};
  config.metrics = [config.metrics];
}
config.metrics.forEach(o => o.name = o.name || name);
```

Also, no need for the `console.log()` since in general we support passing in objects whenever single-item arrays are allowed.

lib/heapwatch.js
```diff
@@ -11,11 +11,12 @@ const ZERO_CUMULATIVE_GC_INTERVAL = {
 const GC_REPORT_INTERVAL = 1000;
 
 class HeapWatch {
-    constructor(conf, logger, statsd) {
+    constructor(conf, logger, statsd, prometheus) {
```
d00rman: Honestly, I don't like this. There should be a more generic way of injecting metrics methods into HeapWatch. A `Metrics` object (or similar) should be given to HeapWatch and have it used. There is really no need for HeapWatch to have to differentiate between statsd and prometheus.

lib/master.js
```diff
@@ -8,6 +8,7 @@ const BaseService = require('./base_service');
 const Worker = require('./worker');
 const Logger = require('./logger');
 const RateLimiterMaster = require('./ratelimiter').master;
+const MetricsEndpoint = require('./prometheus_server');
```
d00rman: Same problem as above. Please create a generic class that is able to handle all types of metrics.

test/utils/simple_config_no_workers.yaml
```diff
@@ -1,6 +1,9 @@
 num_workers: 0
 logging:
   level: fatal
+metrics:
```
d00rman: It would be better to have both prometheus and statsd in the test configs.

lib/heapwatch.js
shdubsh: I took a stab at what you suggested, but I'm not sure you'll like it any better: https://github.com/shdubsh/service-runner/commit/6cd254afec8415ac3dabe77240764aa4835ececf

Let me know?

lib/master.js
shdubsh: This is an HTTP server that exposes the Prometheus metrics endpoint.  I'm not sure what you're asking for here.  Do you think the HTTP server implementation should be in a different file?

lib/worker.js
```diff
@@ -144,14 +141,13 @@ class Worker extends BaseService {
             }, 5000);
         });
 
-        // Metrics reporter
-        this._metrics = makeStatsD(this.config.metrics, this._logger);
-
+        // Metrics reporting
+        this.metrics = makeMetrics(this.config.metrics, this._logger);
```
Pchelolo: Why would we remove the `_` prefix from the internal property name?

lib/metrics/statsd.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    set(value, labels) {
+        labels = formatLabels(this.options, labels);
+        this.client.gauge(normalizeName(labels.join('.')), value);
+    }
+
+    timing(value, labels) {
+        labels = formatLabels(this.options, labels);
+        this.client.timing(normalizeName(labels.join('.')), value);
```
Pchelolo: To avoid code duplication we can probably make a method `_call(methodName, value, labels)` and call it from all the methods.

lib/metrics/index.js
```diff
@@ -0,0 +1,144 @@
...
+            if (o.type === SUPPORTED_CLIENTS.PROMETHEUS) {
+                this.clients.push(new PrometheusClient(o, this.logger));
+            }
+            if (o.type === SUPPORTED_CLIENTS.STATSD) {
+                this.clients.push(new StatsDClient(o, this.logger));
+            }
+            if (o.type === SUPPORTED_CLIENTS.LOG) {
+                o.methods = DEPRECATED_METHODS;
+                this.clients.push(new LogClient(o, this.logger));
+            }
```
Pchelolo: replacing with `switch` would be more readable.

lib/metrics/index.js
```diff
@@ -0,0 +1,144 @@
...
+        this.cache.set(options.name, metric);
+        return metric;
+    }
+
+    fetchMetric(name) {
+        return this.cache.get(name);
+    }
+
+    close() {
+        this.clients.some((o) => o.close());
```
Pchelolo: you mean 'forEach'?

lib/metrics/index.js
```diff
@@ -0,0 +1,144 @@
...
+// }
+    makeMetric(options) {
+        const metric = new Metric(this.clients, this.logger, options);
+        this.cache.set(options.name, metric);
+        return metric;
+    }
+
+    fetchMetric(name) {
+        return this.cache.get(name);
+    }
```
Pchelolo: Why don't we merge the `makeMetric` and `fetchMetric` together, and lazy-create it on demand? What's the point of having 2 sedate methods?

lib/metrics/index.js
```diff
@@ -0,0 +1,144 @@
...
+                this.clients.push(new LogClient(o, this.logger));
+            }
+        });
+    }
+
+    fetchClient(name) {
+        return this.clients.find((client) => {
+            return client.constructor.name === name;
+        });
+    }
```
Pchelolo: Why do we need to expose a 'client'? It seems to be an internal thingy.

lib/metrics/prometheus.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
```
Pchelolo: I believe since 'prometheusMetric' and 'prometheusClient' can only be used together (same as the others), there's no need to put the 'metric' and 'client' in separate files, this can be a private thing within a 'prometheusClient'

lib/metrics/index.js
```diff
@@ -0,0 +1,144 @@
...
+            });
+        };
+    });
+
+    // Deprecated.
+    // Support creating sub-metric clients with a fixed prefix. This is useful
+    // for systematically categorizing metrics per-request, by using a
+    // specific logger.  WARNING: if both Log and StatsD are configured, only
+    // one will be returned.
+    proxy.makeChild = function (name) {
```
Pchelolo: So, what's a replacement for this in a new paradigm?

Pchelolo:
It would be very useful to provide a piece of documentation with a migration guide so that we could get rid of the deprecated stuff sooner rather than later.

lib/worker.js
shdubsh: I suppose there is no good reason to rename it.  I'll fix that up.

lib/metrics/index.js
shdubsh: makeMetric() wants the full object in order to initialize all configured metrics clients.  I figured that providing this object every time you wanted to get one the user already initialized would be onerous.  fetchMetric() is a getter into the cache so that users need not implement their own cache.

lib/metrics/index.js
shdubsh: Our ultimate goal is to eliminate StatsD entirely.  The idea here was to allow users the option to get to the raw client if needed rather than hide the abstraction completely and leave users no other option than to implement it themselves.

lib/metrics/index.js
shdubsh: Yes or using the same metric with a different label value.  More information about the Prometheus data model can be found [here](https://prometheus.io/docs/concepts/data_model/).

In short, the metric names are more static and made multi-dimensional through "tagging" with labels of varying values.

Pchelolo:
ok. This looks fair to me. I think though we shouldn't merge/release it until we have tried this in action in one or two services at least locally.

The minimum requirement for the release would be to test RESTBase/ChangeProp/MCS test with this new version of service-runner to prove backwards compatibility. 

Next, I guess we need to take some small service-template service and try converting it to new metrics...

lib/master.js
```diff
@@ -36,6 +37,7 @@ class Master extends BaseService {
         this._firstWorkerStarted = false;
         this._firstWorkerStartupAttempts = 0;
         this._workerStatusMap = {};
+        this.PrometheusServer = null;
```
d00rman: Nit: please use camelCase

lib/metrics/index.js
```diff
@@ -0,0 +1,147 @@
...
+                case SUPPORTED_CLIENTS.PROMETHEUS:
+                    this.clients.push(new PrometheusClient(o, this.logger));
+                    break;
+                case SUPPORTED_CLIENTS.STATSD:
+                    this.clients.push(new StatsDClient(o, this.logger));
+                    break;
+                case SUPPORTED_CLIENTS.LOG:
+                    this.clients.push(new LogClient(o, this.logger));
+                    break;
+                default:
```
d00rman: `log` is actually the default, let's keep it that way by having it if no spec has been given.

lib/metrics/index.js
d00rman: I think it's reasonable to assume that (i) users want as simplistic as an interface as possible; and (ii) prometheus will not be available in all environments (dev, local, vagrant, etc). In light of that, I would opt for hiding the prometheus logic and client away and provide simple ways for people to send metrics.

lib/metrics/index.js
d00rman: I think too much focus in this PR is given to Prometheus. We should assume sensible defaults and use them. Yes, Prometheus is powerful, but personally I'd want to see more focus on client simplicity and compatibility. If one does need all Prometheus has to offer, it's fine to have a class here that can expose all of that, and then the client can use it and tell service-runner "use this object for metrics", since I figure that 99.99% of clients is perfectly fine with "send metric X with value Y".

lib/metrics/index.js
```diff
@@ -0,0 +1,147 @@
...
+//     },
+//     labels: {
+//         keys: [],
+//         labelPosition: 'before',
+//         omitLabelNames: false
+//     }
+// }
+    makeMetric(options) {
+        const metric = new Metric(this.clients, this.logger, options);
+        this.cache.set(options.name, metric);
```
d00rman: What if the metric already exists with different settings?

shdubsh: If you mean a collision based on name, then it will silently overwrite it in the cache or depending on configuration throw an error from the metrics client itself.  Prometheus names get registered and a collision there will throw an error.  I'm open to suggestions here.

lib/metrics/index.js
shdubsh: Added in latest revision

lib/master.js
shdubsh: Added in latest revision

shdubsh:
There are two main problems that this arrangement attempts to solve:
1. The information required to initialize a Prometheus metric is incompatible both in format and potential options with how StatsD metrics are defined.
2. The interface presented by the StatsD and Prometheus library are very different.  Prometheus has to maintain state so that metrics can be picked up by the collector when it comes around to request it.  In contrast StatsD can fire the metric at the network and immediately forget about it.

I am looking into those tests.  Once I have the results, I will post them here.

shdubsh:
StatsD output compared in service-template-node: https://gist.github.com/shdubsh/6c2cc7ab3c2ae86ad6023897ddd43a61

shdubsh:
Restbase appears to be inconsistent in its reporting of metrics (excluding gc and heap metrics).  There's a 2% difference between baseline Restbase v1.1.4 tests.  When testing this PR against Restbase, the differences between runs is largely the same (~1% improvement).  Compared baseline to this PR is 98% the same.

shdubsh:
Change-propagation appears to be inconsistent in its reporting of metrics (excluding gc and heap metrics).  There's a 6.5% average difference between baseline ChangeProp v0.9.3 test runs.  When testing this PR against ChangeProp, the differences between runs is largely the same (~1.5% regression).  Compared baseline to this PR is 97% the same.

shdubsh:
Mobileapps appears to be consistent in its reporting of metrics (excluding gc and heap metrics).  There is no difference between this PR and mobileapps/master.

lib/metrics/log.js
```diff
@@ -0,0 +1,82 @@
...
+        this.timing(Date.now() - startTime, labels);
+    }
+}
+
+// A simple console reporter. Useful for development.
+class LogClient {
+    constructor(options, logger) {
+        this._logger = logger;
+        this._serviceName = options.name ? `${options.name}.` : 'service-runner.';
+        this.methods = options.methods;
```
ottomata: What is methods supposed to be for log client?

ottomata:
Can we merge this into a experimental service-runner branch and use it in wmf prod?  I'm working on porting eventstreams to k8s this quarter, and am having a bit of trouble getting the statsd prometheus mappings to work.  I'd much rather not worry about statsd at all and just use prometheus.  I'm currently running eventstreams service-runner at this PR, and things seem to work mostly ok.

lib/metrics/log.js
shdubsh: Great catch!  Somehow this piece got missed.  I just pushed a fix.

lib/heapwatch.js
```diff
@@ -47,16 +47,26 @@ class HeapWatch {
...
             this._gcStats.on('stats', this.reportStatsHandler);
             this.gcReportInterval = setInterval(() => {
-                const timings = {};
+                const gcMetrics = this.metrics.makeMetric({
+                    type: 'Gauge',
+                    name: 'gc',
+                    prometheus: {
+                        name: 'service_runner_heapwatch_gc_ns',
+                        help: 'heapwatch gc pause ns',
+                        staticLabels: { service: this.metrics.getServiceName() }
```
ottomata: Should this also be added automatically for other metrics created via service runner?  E.g. if I do app.metrics.makeMetric and use promtheus, perhaps this staticLabel; should somehow be auto added?  Maybe a app.makeMetric proxy?

shdubsh: Based only on the [k8s citoid config](https://github.com/wikimedia/operations-deployment-charts/blob/master/charts/citoid/config/prometheus-statsd.conf), it appears this particular static label is being set for heapwatch and request duration metrics.  For Hyperswitch users, we intend to add this staticLabel in its request handler, but for projects based on service-template-node, they will likely have to set this attribute in their request handler.

Pchelolo:
This looks good to me. There's a lot, so I guess we'd missed some bits and pieces, but I guess we can followup on those once we start moving it towards production.

Step 1 would be to choose a canary. Let's maybe start from a non-k8s-based service so that the statsd exporter is not complicating things. I propose mobileapps  (cc @berndsi ) - it is well understood, it exists in beta, it has a solid graphing dashboard. 

Since it's still a risky operation, I propose to deploy it as follows:
- change the target of this PR to 'new_metric' branch
- for service-template-node PR we change the target to a branch, merge it. also change the service-runner dependency to the git dependency.
- for mobileapps we follow https://www.mediawiki.org/wiki/ServiceTemplateNode/Updating to create a patch for mobileapps. 
- we deploy that to beta and see that it works with statsd
- we make a change for beta-only for mobileapps deploy repo
- we deploy that to beta make the dashboard in beta based on https://grafana.wikimedia.org/d/000000183/mobileapps?orgId=1 but on prometheus. 
- if all good, we continue to prod. 
- if all good, we select the next canary based on k8s. After 2 canaries we should release the package and convert everything. But that point we will be more confident, so we can move faster.

We need to coordinate all this with reading infrastructure. Please ping me on IRC for coordination, I guess we better begin on Monday if you have time to do all the preparations by then.

shdubsh:
@Pchelolo This plan looks good to me.  I do not have access to this repository to create the target branch.  If someone would create this branch, I'm happy to re-target this PR.

Pchelolo:
you got a 'prometheus_metrics' branch now.

berndsi:
+ Product Infrastructure since I'll be out until sometime in February.

ottomata:
Oh, hey, can I be the guinea pig canary?

I'm already in the process of [helmifying eventstreams](https://gerrit.wikimedia.org/r/c/operations/deployment-charts/+/551843), and I'm [using this service-runner PR.](https://gerrit.wikimedia.org/r/c/mediawiki/services/eventstreams/+/559568)

I'll see if I can update with service-template-node stuff as Petr suggests.

ottomata:
BTW, I just created a prometheus_metrics branch from master in service-template-node too.  @shdubsh you should be able to point [your PR](https://github.com/wikimedia/service-runner/pull/221) there at it, then we can merge them both into the branches!

ottomata:
I want to rebuild eventstreams using this from the branch instead of this PR, so I'm merging this into the new branch now.  Hope you don't mind, Petr! :)

PR #223, opened 2020-02-11
Update dependencies and release 2.7.4

Pchelolo:
Removing node 6 from travis since it's no longer supported by eslint and is almost eol for us as well


.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "6"
```
clarakosi: For htcp-purge we used Node 8, 10, and 12... should we add that here? Also, do we want to move this away from Travis to Github Actions?

Pchelolo: hm... regarding 8 - we don't really use or plan on using 8, so.. meh.. regarding gh actions.. why not.

Pchelolo:
done done and done

Pchelolo:
published.

mdholloway:
o/ A bunch of services (mobileapps, recommendation-api, proton?) are still stuck on Node 6 until we can get them onto the deployment pipeline.  Can we keep yargs at 14.2.0 for now?  (Node 6 support is dropped in 15.0.0.)

Pchelolo:
Hm. Oh ye, didn’t notice that. Will return 6 support

PR #225, opened 2020-03-06
Update heapwatch metrics

shdubsh:
Update heapwatch metric names and label usage per feedback in T238658

Bug: T238658

### DIFF
--- a/lib/heapwatch.js
+++ b/lib/heapwatch.js
@@ -47,23 +47,19 @@ class HeapWatch {
             this._gcStats = require('gc-stats')();
             this._gcStats.on('stats', this.reportStatsHandler);
             this.gcReportInterval = setInterval(() => {
-                const gcMetrics = this.metrics.makeMetric({
-                    type: 'Gauge',
-                    name: 'gc',
-                    prometheus: {
-                        name: 'nodejs_gc_duration_ns',
-                        help: 'garbage collection pause duration ns',
-                        staticLabels: { service: this.metrics.getServiceName() }
-                    },
-                    labels: {
-                        names: ['type'],
-                        omitLabelNames: true
-                    }
-                });
                 Object.keys(this.cumulativeGCTimes).forEach((gcType) => {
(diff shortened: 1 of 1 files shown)


lib/heapwatch.js
```diff
@@ -47,23 +47,19 @@ class HeapWatch {
...
                 Object.keys(this.cumulativeGCTimes).forEach((gcType) => {
                     const totalGCTime = this.cumulativeGCTimes[gcType];
                     if (totalGCTime > 0) {
-                        gcMetrics.timing(totalGCTime, [gcType]);
+                        this.metrics.makeMetric({
+                            type: 'Histogram',
+                            name: `gc.${gcType}`,
+                            prometheus: {
+                                name: `nodejs_gc_${gcType}_duration_ns`,
+                                help: 'garbage collection pause duration ns',
```
filippog: We should use seconds as a base unit

lib/heapwatch.js
```diff
@@ -47,23 +47,19 @@ class HeapWatch {
...
+                                staticLabels: { service: this.metrics.getServiceName() },
```
filippog: AFAICT the GC and heap metrics are scoped to the whole nodejs runtime and not per-service (?) (internal to service runner), in this case we shouldn't attach `service` label. cc @Pchelolo to confirm/invalidate my assumption above

shdubsh: The service name is populated by inspecting [package.json](https://github.com/wikimedia/service-runner/blob/prometheus_metrics/lib/base_service.js#L355) for the name field and defaulting to "service-runner."  From the metrics perspective, this name can be [overridden by adding a name field to the config.metrics[] member object.](https://github.com/wikimedia/service-runner/blob/prometheus_metrics/lib/base_service.js#L399)

Service name from the metrics perspective is populated [here](https://github.com/wikimedia/service-runner/blob/prometheus_metrics/lib/base_service.js#L383-L400).

ottomata: Ya, and the service label is used to pull together most dashboards, so we should keep it.

PR #226, opened 2020-03-30
Prefixing and Normalizing Prometheus metrics

shdubsh:
Add default prefix to Prometheus metrics.
Normalize Prometheus metric name and label keys and values.

Bug: T247820


shdubsh:
Closing in response to discussion.  The global nature of this change does not align to what we're hoping to accomplish.

PR #227, opened 2020-03-31
Selectively toggle service label based on environment variable.

shdubsh:
Add env variable toggle to selectively disable service label.
Normalize Prometheus metric name and label keys and values.

Bug: T247820

### DIFF
--- a/lib/metrics/index.js
+++ b/lib/metrics/index.js
@@ -53,6 +53,16 @@ class Metrics {
         return this.serviceName;
     }
 
+    // T247820: Selectively disable service label based on environment variable.
+    // Intended for production use case where the service label will be set
+    // by the Prometheus server.
+    getServiceLabel() {
+        if (process.env.METRICS_SERVICE_LABEL_ENABLED === 'false') {
+            return {};
+        }
+        return { service: this.serviceName };
+    }
+
     fetchClient(name) {
         return this.clients.find((client) => {
             return client.constructor.name === name;
--- a/lib/metrics/prometheus.js
+++ b/lib/metrics/prometheus.js
@@ -17,6 +17,9 @@ class PrometheusMetric {
                 });
             }
         }
+        options.prometheus.name = this._normalize(options.prometheus.name);
+        options.prometheus.labelNames = options.prometheus.labelNames || [];
+        options.prometheus.labelNames = options.prometheus.labelNames.map(this._normalize);
         this.metric = new this.client[options.type]({
             name: options.prometheus.name,
             help: options.prometheus.help,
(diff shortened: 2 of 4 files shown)


lib/metrics/index.js
```diff
@@ -53,6 +53,16 @@ class Metrics {
         return this.serviceName;
     }
 
+    // T247820: Selectively disable service label based on environment variable.
+    // Intended for production use case where the service label will be set
+    // by the Prometheus server.
+    getServiceLabel() {
+        if (process.env.DISABLE_SERVICE_LABEL === 'true') {
```
ottomata, line 60: I think its much better to keep switches like this named positivliy.  `disabled = true` is not as understandablle as `enabled = false`.  Can we call this `METRICS_SERVICE_LABEL_ENABLED` and default to true?

clarakosi:
Adding @nikkhn and @Pchelolo to pair review this pull request

lib/metrics/prometheus.js
```diff
@@ -34,23 +37,33 @@ class PrometheusMetric {
         }
     }
 
+    _normalize(str) {
+        return String(str).replace( /[^a-z0-9]/g, '_' ) // replace non-alphanumerics
```
Pchelolo, line 41: No need to new String. The string is a String.

Pchelolo: alphanumeric regex in JS is `\w`

lib/metrics/prometheus.js
```diff
@@ -34,23 +37,33 @@ class PrometheusMetric {
...
+
     increment(amount, labels) {
         this._handleStaticLabels(labels);
+        labels.map(this._normalize);
         this.metric.labels.apply(this.metric, labels).inc(amount);
     }
 
     decrement(amount, labels) {
         this._handleStaticLabels(labels);
+        labels.map(this._normalize);
```
Pchelolo, line 54: You're not assigning the result to labels. .map does not modify the array itself afaik.

lib/metrics/index.js
```diff
@@ -53,6 +53,16 @@ class Metrics {
         return this.serviceName;
     }
 
+    // T247820: Selectively disable service label based on environment variable.
+    // Intended for production use case where the service label will be set
+    // by the Prometheus server.
+    getServiceLabel() {
```
Pchelolo, line 59: Perhaps we should handle that in prometheus _handleStaticLabels? The users of the metrics API can still set the `service` label and as I understand the idea is to drop it right?

shdubsh: It is written this way to be as flexible as possible.  For users outside of WMF production, the service label can be whatever the user wants, including the option to get the value provided by `getServiceLabel()`.  For WMF Production, this label will be set by the Prometheus Server at scrape time.

The toggle is specifically about whether or not to provide the label in heap metrics.  The service label in heap metrics was [defined a while back](https://github.com/wikimedia/operations-deployment-charts/blob/master/charts/citoid/config/prometheus-statsd.conf#L18), but developers using service-runner expressed interest in retaining the service label present by default in the metrics endpoint.

To your question, no, the intent is not to drop the service label entirely.  Users of service-runner can still set the service label to whatever they like for any metric they define.  For instances in WMF Production, this label will be externally set.

lib/metrics/prometheus.js
shdubsh: From tests, it appears str is not String() (nodejs 13):
```
Uncaught TypeError: str.replace is not a function
      at _normalize (lib/metrics/prometheus.js:41:20)
```

PR #228, opened 2020-04-08
Add support for having named logging levels

akosiaris:
Why:
service-runner, when logging to stdout delegates to bunyan stdout
logging process that ends up creating log entries with level specified
as an integer, even though the code everywhere might be using named
levels. This hasn't been a problem up to now because we sparringly
relied on that, but with the move to service-runner apps on kubernetes
and the logging pipeline this is starting to cause issues, namely
https://phabricator.wikimedia.org/T239459. It would be nice if
service-runner allowed to log to stdout named loglevels

What:
Add a simple, albeit probably inefficient function due to all the JSON
encoding/decoding that wraps process.stdout and changes the level
attribute from an integer to a string. Provide it as an additional
stream.

The output looks like this:

{"name":"myapp","hostname":"buster","pid":5835,"level":"INFO","msg":"This is logging","time":"2020-04-08T09:03:09.294Z","v":0}

instead of:

{"name":"myapp","hostname":"buster","pid":5835,"level":30,"msg":"This is logging","time":"2020-04-08T09:03:09.294Z","v":0}

Bump service-runner version to reflect the change

Bug: T239459

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -12,6 +12,17 @@ function extractSimpleLevel(levelPath) {
     return typeof levelPath === 'string' && levelPath.split('/')[0];
 }
 
+function NamedLevelStdout() {
+    return {
+        write: (entry) => {
+            // Kind of inefficient approach
+            var logObject = JSON.parse(entry);
+            logObject.level = bunyan.nameFromLevel[logObject.level].toUpperCase();
+            process.stdout.write(JSON.stringify(logObject) + '\n');
+        }
+    };
+}
+
 const streamConverter = {
     gelf(stream, conf) {
         let host = stream.host;
@@ -43,6 +54,12 @@ const streamConverter = {
     stdout(stream, conf) {
         return {
             stream: process.stdout,
+            level: stream.level || conf.level
+        };
+    },
+    namedstdout(stream, conf) {
+        return {
+            stream: NamedLevelStdout(),
             level: stream.level || conf.level
         };
     },
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const Writable = require('stream').Writable;
 const bunyan = require('bunyan');
 const gelfStream = require('gelf-stream');
 const syslogStream = require('bunyan-syslog-udp');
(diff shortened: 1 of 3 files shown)


akosiaris:
I fail to see how I can debug the failed tests. I see that `npm test` failed, but I get a 

`We are currently unable to download the log. Please try again later.` when I try to see why. I already waited 2 hours, did this thing work in the past?

clarakosi:
Hmm, maybe the logs were delayed? I see it now when I go to the details[1] and also in the actual files[2].

You are missing parenthesis around `entry` in line 17 of logger.js

1. https://github.com/wikimedia/service-runner/pull/228/checks?check_run_id=570240867
2. https://github.com/wikimedia/service-runner/pull/228/files

akosiaris:
@clarakosi, sigh no it turns out my browser was blocking the XHR to fetch those. But thanks! I 'll fix that shortly :)

Pchelolo:
So, a few comments:
1. Why not make the logging pipeline understand these levels?
2. In order to avoid JSON decoding/encoding, you need to define your stream as 'raw' https://github.com/trentm/node-bunyan#stream-type-raw - then it will get non-encoded objects passed.
3. The stream implementation should be a class extending node native Writable stream and overriding '_write' method and created in an object mode. Right now you're abusing node dynamic nature. I can help with that if you want, it would be faster for me to write it then to explain what should be done :)
4. There's no way you'd need both named and default stdout stream. Should we just support a property in stdout stream configuration that will name the levels instead of building a new stream?

akosiaris:
> 1. Why not make the logging pipeline understand these levels?

That would make the pipeline aware of unique to `bunyan` internals for which there is no guarantee there will be no change in semantics, format, type etc at some point in the future.

> 2. In order to avoid JSON decoding/encoding, you need to define your stream as 'raw' https://github.com/trentm/node-bunyan#stream-type-raw - then it will get non-encoded objects passed.

Oh, nice. I did not know that.

> 3. The stream implementation should be a class extending node native Writable stream and overriding '_write' method and created in an object mode. Right now you're abusing node dynamic nature. I can help with that if you want, it would be faster for me to write it then to explain what should be done :)

I 'll admit I only have a vague understanding of what that entails. If you feel it will easier for you to just write it, please feel free to.

> There's no way you'd need both named and default stdout stream. Should we just support a property in stdout stream configuration that will name the levels instead of building a new stream?

I supported both ways just to ease the migration of services from the default to the named format.  But if we can just have a property the achieves the same thing, even better.

Pchelolo:
> I 'll admit I only have a vague understanding of what that entails. If you feel it will easier for you to just write it, please feel free to.

Done :) 

Given that I mostly rewrote your code, I would like to reverse the review now, could you review my code now? :)

lib/logger.js
```diff
@@ -41,8 +61,22 @@ const streamConverter = {
...
+                level: stream.level || conf.level
+            };
+        } else {
+            return {
+                stream: process.stdout,
+                level: stream.level || conf.level
+            };
+        }
+    },
+    namedstdout(stream, conf) {
```
akosiaris, line 77: I guess we no longer need this function?

akosiaris:
> Given that I mostly rewrote your code, I would like to reverse the review now, could you review my code now? :)

Hah, not sure how good of a review you can expect from me, but I did have a look, I think we need to remove my initial namedstout() function, but other LGTM

lib/logger.js
Pchelolo: Of course

PR #232, opened 2020-10-22
Prom metrics

clarakosi:


lib/metrics/prometheus.js
```diff
@@ -32,7 +32,10 @@ class PrometheusMetric {
     _handleStaticLabels(labels) {
         if (this.staticLabels !== undefined) {
             Object.keys(this.staticLabels).forEach((name) => {
-                labels.unshift(this.staticLabels[name]);
+                const normalizedLabel = this._normalize(this.staticLabels[name]);
```
Pchelolo, line 35: Can you move this normalization of static labels into constructor so that you do it once?

e.g. this.staticLabels = options.prometheus.staticLabels.map(this._normalize) or something?

PR #233, opened 2020-12-08
Fix label modifications in statsd and prometheus

clarakosi:


lib/metrics/prometheus.js
```diff
@@ -47,49 +49,56 @@ class PrometheusMetric {
     }
 
     increment(amount, labels) {
-        this._handleStaticLabels(labels);
-        labels = labels.map(this._normalize);
-        this.metric.labels.apply(this.metric, labels).inc(amount);
+        let updatedLabels = labels;
```
Pchelolo, line 52: Javascript doesn't create a copy on assignment. This is still pointing to exact same array. This patch changes nothing...

You need to actually clone the labels array... let updatedLabels = [...labels] will work. Just check this syntax works with node 6. Alternatively, let updatedLabels = labels.map(label => label) should work too.

lib/metrics/prometheus.js
```diff
@@ -6,27 +6,29 @@ const Prometheus = require('prom-client');
...
+        if (this.options.labels === undefined) {
+            this.options.labels = { names: [] };
         }
-        this.staticLabels = options.prometheus.staticLabels;
+        this.staticLabels = this.options.prometheus.staticLabels;
         if (this.staticLabels !== undefined) {
             if (Object.keys(this.staticLabels).length > 0) {
                 Object.keys(this.staticLabels).forEach((name) => {
-                    options.labels.names.unshift(name);
+                    this.options.labels.names.unshift(name);
```
Pchelolo, line 17: This actually modifies the original labels and original options still. Everything is passed/assigned by reference.

Pchelolo: You can depend on https://www.npmjs.com/package/lodash.clonedeep and use it to create deep copies where needed. Like here it will be this.options = cloneDeep(options) and then you modify this.options like you did.

clarakosi: Updated and tested with node 6

lib/metrics/prometheus.js
```diff
@@ -47,49 +50,56 @@ class PrometheusMetric {
     }
 
     increment(amount, labels) {
-        this._handleStaticLabels(labels);
-        labels = labels.map(this._normalize);
-        this.metric.labels.apply(this.metric, labels).inc(amount);
+        let updatedLabels = [...labels];
+        this._handleStaticLabels(updatedLabels);
```
Pchelolo, line 54: Would actually be nicer if you clone them inside handleStaticLabels and return new labels.

then here you would be like `const updatedLabels = this._handleStaticLabels( labels );`. More readable.

Pchelolo (changes requested):
One more nitpick

lib/metrics/prometheus.js
```diff
@@ -47,49 +52,49 @@ class PrometheusMetric {
     }
 
     increment(amount, labels) {
-        this._handleStaticLabels(labels);
-        labels = labels.map(this._normalize);
-        this.metric.labels.apply(this.metric, labels).inc(amount);
+        let updatedLabels = this._handleStaticLabels(labels);
```
Pchelolo, line 55: const. it can be const now.

Pchelolo (changes requested):
one more nit pick

lib/metrics/prometheus.js
clarakosi: But we are re-assigning it in the next line

Pchelolo: Can you merge the lines? 

```
const updatedLabels = this._handleStaticLabels(labels).map(this._normalize);
```
?

Pchelolo: const is always much better then let.

PR #234, opened 2021-01-11
Remove logging when no metrics are passed in

clarakosi:


lib/base_service.js
```diff
@@ -394,8 +394,7 @@ class BaseService {
 
             // And metrics clients, massaging it into an array
             if (!Array.isArray(config.metrics)) {
-                config.metrics = config.metrics || {};
-                config.metrics = [config.metrics];
+                config.metrics = Object.keys(config.metrics).length === 0 ? [] : [config.metrics];
```
Pchelolo, line 397: Object.keys(undefined) will produce a TypeError. Why do you need Object.keys here?

clarakosi: I wanted to confirm if the object was empty since it was being defaulted to an empty object as part of `_sanitizeConfig()`

Pchelolo: Do `(!config.metrics || !Object.keys(config.metrics).length)` just to be sure

clarakosi: 👍

PR #235, opened 2021-03-31
Bugfix: clone labels when interpolating metric name

shdubsh:
When the log metrics output is first in line,
`metrics.utils.formatLabels` will mutate the labels array.
This mutation will leak to the other metrics clients and manifest
as an Invalid number of arguments in prom-client.

Bug: T278141


Pchelolo:
Makes sense to me, please bump a version. Will leave to @clarakosi to have a look, she was working on this code recently.

clarakosi (approved):

PR #236, opened 2021-04-15
Drop Node 6 support and bump to 3.0.0

jdforrester:
Fixes https://phabricator.wikimedia.org/T240727
Addresses https://github.blog/2021-09-08-github-security-update-vulnerabilities-tar-npmcli-arborist/


.github/workflows/nodejs.yml
```diff
@@ -9,13 +9,13 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [6.x, 10.x, 12.x]
+        node-version: [10.x, 12.x]
```
Pchelolo, line 12: Yeah, that's a thing. There's still some stuff running 6 :(

jdforrester: Sure, but they won't be getting new versions of this.

Pchelolo: in that case, please bump a version of the package. Practically this means dropping node version support, so maybe let's respect server and go major?

jdforrester: Sure, done.

sbassett29:
So the tests are ostensibly failing on npm not liking `gc-stats@1.5.0`.  Assuming we want to keep the optional dependency ([do we?](https://phabricator.wikimedia.org/T290618#7353360)) could we just version it at `github:dainis/node-gcstats#fd4ea294c9a0eda05a760348424df2ce072659ea` as `npm i` builds it in package-lock.json?  There's [some precedent](https://gerrit.wikimedia.org/g/mediawiki/libs/Dodo/+/1ad980abf89712bad61a0e600efe1c2cfc79c0a2/package-lock.json) for doing this in production code.

Pchelolo:
As I indicated on the ticket I think we should just drop the feature entirely.

package.json
```diff
@@ -48,14 +48,14 @@
     "yargs": "^14.2.2"
   },
   "optionalDependencies": {
-    "gc-stats": "^1.4.0",
+    "gc-stats": "dainis/node-gcstats#fd4ea294c9a0eda05a760348424df2ce072659ea",
```
Pchelolo, line 51: If #240 lands this will not be needed.

Pchelolo (approved):
Needs a rebase, afterwards we can release this.

jdforrester:
> As I indicated on the ticket I think we should just drop the feature entirely.

Done (I think).

Pchelolo:
haha, great minds think alike #240

jdforrester:
I'll rebase mine onto yours, much simpler.

PR #238, opened 2021-07-07
Don't normalize prometheus label values

ottomata:
Slight refactor of prometheus label names initialization.  Should
be a no-op that just makes the code a bit more readable.


shdubsh (approved):
LGTM, Thanks!

ottomata:
@Pchelolo realized we need this too.  Can we merge this and make a new service-runner release (2.8.4?) and then I can patch service-template-node to use that?

Pchelolo:
Released and published.

ottomata:
Thank you!

PR #240, opened 2021-09-15
Remove GC stats feature

Pchelolo:
The feature emits metrics about garbage collection.
To my knowledge these metrics have never benefitted
anyone in 5 years of their existance. Now gc-stats
package used to collect the metric has issues, so
let's just remove the feature entirely.


Pchelolo:
cc @sbassett29

jdforrester (approved):

PR #243, opened 2021-11-09
build: Upgrade eslint-config-wikimedia to 0.21.0

jdforrester:
Auto-fix existence checks via `.indexOf()` into `.includes()`.

### DIFF
--- a/lib/base_service.js
+++ b/lib/base_service.js
@@ -234,10 +234,10 @@ class BaseService {
         .argv;
 
         args.deployRepo = args.deployRepo || args.review;
-        args.build = args._.indexOf('build') !== -1 || args.deployRepo;
-        args.dockerStart = args._.indexOf('docker-start') !== -1;
-        args.dockerTest = args._.indexOf('docker-test') !== -1;
-        args.generate = args._.indexOf('generate') !== -1;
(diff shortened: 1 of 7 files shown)


Pchelolo:
Seem to need node 12. We're still running a lot of node 10 in production though.. To soon?

jdforrester:
> Seem to need node 12. We're still running a lot of node 10 in production though.. To soon?

Migration is sort-of underway at https://phabricator.wikimedia.org/T290750 – I think it's reasonable to get this done soon. But we haven't got v3.0.0 into https://github.com/wikimedia/service-template-node yet. :-(

mvolz:
@jdforrester, is there anything left to do in this PR?

jdforrester:
> @jdforrester, is there anything left to do in this PR?

Theoretically we're waiting for all prod services to move so we don't have a situation with an incident where we can't release a security update because some services haven't switched yet.

jdforrester:
> > @jdforrester, is there anything left to do in this PR?
> 
> Theoretically we're waiting for all prod services to move so we don't have a situation with an incident where we can't release a security update because some services haven't switched yet.

At this point, let's just ship it?

PR #244, opened 2021-11-24
Switch prometheus timings to seconds

akosiaris:
Date.now() returns milliseconds and startTime is assumed to be a
Date.now() call result too.

In prometheus land, the rule is to always use seconds[1] and not have to
care about the prefixes of units[2]

The vast majority of metric names out there follow that pattern. Instead
of diverging, causing confusions to developers who need to adapt their
heads around naming metric differently as well as setting histogram
buckets differently, conform to the status quo and divide by 1000.
Precision won't be lost as we just move around the decimal separator.

This is a breaking change for applications/services that have migrated
to prometheus, so bump to 3.1.0. Charts and statistics might have to be
updated as well as dashboards.

[1] https://prometheus.io/docs/instrumenting/writing_exporters/#naming
[2] https://www.robustperception.io/who-wants-seconds

Signed-off-by: Alexandros Kosiaris <akosiaris@wikimedia.org>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,5343 @@
 {
   "name": "service-runner",
-  "version": "3.0.0",
+  "version": "3.1.0",
   "lockfileVersion": 1,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "service-runner",
+      "version": "3.1.0",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "bluebird": "^3.7.2",
+        "bunyan": "^1.8.12",
+        "bunyan-syslog-udp": "^0.2.0",
+        "dnscache": "^1.0.2",
+        "gelf-stream": "^1.1.1",
+        "hot-shots": "^6.8.7",
+        "js-yaml": "^3.13.1",
+        "limitation": "^0.2.3",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1,8 +1,5343 @@
 {
   "name": "service-runner",
-  "version": "3.0.0",
-  "lockfileVersion": 1,
```
mszabo-wikia, line 4: Unintended change? This may confuse Node <= 10

akosiaris: Definitely unintended. I don't even know what that does. 

To be honest, I am a bit wondering why we commit this file in the git repo. I understand why we would that for anything that is to be deployed, but does this argument also stand for a library (which service-runner is)?

Anyway, I 'll amend.

akosiaris:
Sigh, I made a blunder... I 'll revert and restart this. Sorry about that.

Pchelolo:
This would've been quite easy to review though :)

package-lock.json
mszabo-wikia: I think this is still important for reproducible builds e.g. in CI and during publish. Otherwise, we may get different resolutions for each run

PR #245, opened 2021-11-24
Switch prometheus timings to seconds #2

akosiaris:
This effectively reverts commit
844fe44bb79b08bbc307534ddc72399c46776158, whose parent was mistakenly
pushed instead of being merged as a PR. Below is the original PR

Date.now() returns milliseconds and startTime is assumed to be a
Date.now() call result too.

In prometheus land, the rule is to always use seconds[1] and not have to
care about the prefixes of units[2]

The vast majority of metric names out there follow that pattern. Instead
of diverging, causing confusions to developers who need to adapt their
heads around naming metric differently as well as setting histogram
buckets differently, conform to the status quo and divide by 1000.
Precision won't be lost as we just move around the decimal separator.

This is a breaking change for applications/services that have migrated
to prometheus, so bump to 3.1.0. Charts and statistics might have to be
updated as well as dashboards.

[1] https://prometheus.io/docs/instrumenting/writing_exporters/#naming
[2] https://www.robustperception.io/who-wants-seconds

### DIFF
--- a/lib/metrics/prometheus.js
+++ b/lib/metrics/prometheus.js
@@ -98,7 +98,7 @@ class PrometheusMetric {
 
     endTiming(startTime, labelValues) {
         const updatedLabelValues = this._getLabelValues(labelValues);
-        this.timing(Date.now() - startTime, updatedLabelValues);
+        this.timing((Date.now() - startTime) / 1000, updatedLabelValues);
     }
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,5343 @@
 {
   "name": "service-runner",
-  "version": "3.0.0",
+  "version": "3.1.0",
   "lockfileVersion": 1,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "service-runner",
+      "version": "3.1.0",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "bluebird": "^3.7.2",
+        "bunyan": "^1.8.12",
+        "bunyan-syslog-udp": "^0.2.0",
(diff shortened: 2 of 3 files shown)


akosiaris:
Let's try to do #244 correctly now :-)

Pchelolo:
Makes sense to me. I think event services maintained by @ottomata are the major users of native prometheus, so I'll let him have a chance to say something

mszabo-wikia:
Would it make sense to backport this to v2.8.x? I see some of the services are not yet on v3 ;d

akosiaris:
> Would it make sense to backport this to v2.8.x? I see some of the services are not yet on v3 ;d

I don't think there is such a versioning policy (e.g. supporting a prior version), at least I 've never read or heard about any. If there is, we could do that, but otherwise, just bump projects to v3 ?

mszabo-wikia:
@akosiaris yeah, but it requires migrating metrics invocations to the new API, which seems to be the main reason why most/all wikimedia node services remain on 2.8.x

akosiaris:
@ottomata gave his thumbs up at https://phabricator.wikimedia.org/T222795#7578389 so this can be merged.

akosiaris:
> @akosiaris yeah, but it requires migrating metrics invocations to the new API, which seems to be the main reason why most/all wikimedia node services remain on 2.8.x

Could be, but on the other both 2.9 and 3.0 were released on Sept 15th. My money is more on the "most wikimedia nodejs services aren't even aware that 3.0 exists" than on the "they will stall upgrading this because it might require some work".

PR #252, opened 2024-02-27
Upgrade prom-client from ^11.5.3 to ^12.0.0

jdforrester:
12.0.0 provides GC metrics which are wanted for various uses.

The breaking changes are mostly the dropping of support for Node 6 and 8, which don't affect us, removal of deprecated constructor use, and change in how default metrics are now collected, which aren't currently used:

  https://github.com/siimon/prom-client/releases/tag/v12.0.0

https://phabricator.wikimedia.org/T350180

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
 		"js-yaml": "^3.13.1",
 		"limitation": "^0.2.3",
 		"lodash.clonedeep": "^4.5.0",
-		"prom-client": "^11.5.3",
+		"prom-client": "^12.0.0",
 		"semver": "^7.6.0",
 		"tar": "^6.2.0",
 		"yargs": "^17.7.2"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
 		"js-yaml": "^3.13.1",
 		"limitation": "^0.2.3",
 		"lodash.clonedeep": "^4.5.0",
-		"prom-client": "^11.5.3",
+		"prom-client": "^12.0.0",
```
gmodena, line 46: The latest version is 15.1.0. Should we be bold and aim for that?

gmodena (approved):
LGTM. 

Left a comment re which version of prom-client to pin, but 12.x has support for the GC metrics we'd like to report.

package.json
jdforrester: There are a bunch of other breaking changes, and no tests, so I was trying to be conservative in the upgrade.

PR #253, opened 2024-03-05
tests: Replace use of preq with native fetch

jdforrester:
Bug: T293853
Bug: T309772

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,8 +33,7 @@
 				"eslint-config-wikimedia": "0.28.2",
 				"mocha": "^10.2.0",
 				"mocha-lcov-reporter": "^1.3.0",
-				"nyc": "^15.1.0",
-				"preq": "^0.5.14"
+				"nyc": "^15.1.0"
 			},
 			"engines": {
 				"node": ">=20"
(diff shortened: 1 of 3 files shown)


santhoshtr:
If we increase the node version to 18(LTS, Maintainance mode)  or later, 'fetch' API is standardized and stable in node. No need for an external library. 
(cxserver just migrated away from preq using node native fetch https://phabricator.wikimedia.org/T350773)

jdforrester:
> If we increase the node version to 18(LTS, Maintainance mode) or later, 'fetch' API is standardized and stable in node. No need for an external library. (cxserver just migrated away from preq using node native fetch https://phabricator.wikimedia.org/T350773)

Yes, but other parts of the library aren't yet compatible with Node 18, so we can't do that right now (until I or someone else has the time to fix those incompatibilities).

mvolz:
Not sure what's causing the errors here, but as I'm looking into preq for an unrelated issue, I thought I should note it differs from fetch in at least two ways: 

Preq is wrapping errors, so we might loose some logging features upstream if those are not replaced upstream; 

Preq uses another defunct library called requestretry to automatically retry requests once after a short time out.

mvolz:
Weird this works in 20 but not 18 now. Should we just have this be 7.0.0. and drop 18? :)

jdforrester:
> Weird this works in 20 but not 18 now. Should we just have this be 7.0.0. and drop 18? :)

Yeah, I'm not quite sure what's going on there – possibly an upstream library version issue? Anyway, let's do that (but in a different PR); most users have bumped by now.

mvolz (approved):

PR #255, opened 2024-03-15
wip: enable gc metric collection

gmodena:
I am opening this WIP PR to validate direction and collect feedback on an approach to enable prometheus default metrics collection.

Happy to drop / rebase atop other work we might have in flight.

Implementation details have been discussed in:

Bug: https://phabricator.wikimedia.org/T350180

PR #256, opened 2024-05-04
Enable gc metric collection in Prometheus

jdforrester:
Fork of #255 so I can edit it too.


jdforrester:
OK! This is now mergeable. @gmodena, if you're OK I'll land this?

PR #260, opened 2024-10-22
Require node 18 or higher, update to 6.0.0

mvolz:
Due to localhost now pointing to IPv6 instead of IPv4 as of node 18, TestServer now explicitly serves 127.0.0.1 instead of localhost. 

Consumers who import the TestServer from service runner will need to update their test runner applications to serve from 127.0.0.1 as well in order to upgrade from node16 to node 18, in addition to upgrading from 5.0.0 to 6.0.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "service-runner",
-	"version": "5.0.0",
+	"version": "6.0.0",
 	"description": "Generic nodejs service supervisor / cluster runner",
 	"main": "service-runner.js",
 	"bin": {
@@ -31,7 +31,7 @@
 	},
 	"homepage": "https://github.com/wikimedia/service-runner",
 	"engines": {
-		"node": ">=16"
+		"node": ">=18"
 	},
 	"dependencies": {
 		"bluebird": "^3.7.2",


jdforrester:
The breaking change part of this PR is OK, but tagging this particular patch as the 6.0.0 release seems premature?

mvolz:
Better? :)

jdforrester:
> Better? :)

No, worse. This is a breaking change, so the next release will definitely be 6.0.0. I just don't think this MR should be considered the release point.

mvolz:
This release point includes this breaking change: https://github.com/wikimedia/service-runner/commit/f4f793f1e3bf96c7fe0be2b1a5293561eedc402c

I would like to release it ASAP as it's blocking us from updating citoid to 18...20.

edg2s (approved):

edg2s:
What else needs to be in this release? I'm happy for this to go as is...

jdforrester:
> What else needs to be in this release? I'm happy for this to go as is...

If we're doing this, we could land !253 too, for instance. But eh, fine, let's just call this the release.

PR #264, opened 2024-12-19
(Actually) suppress yargs warning

mvolz:
Follow up to 1b7ce8d41394a

Move suppression command up earlier to suppress
all warnings.

Increment patch version.


mvolz:
Didn't do my due diligence when reviewing this before, sorry! Anyway this works.

jdforrester (approved):
Huh.

